Issue #19, opened 2017-11-11
Create-league/tournament page


MarcelSchmidberger:
**AS A** user after login
**I WANT TO** be able to create a new league or tournament
**SO THAT I** start a new sport event

**ACCEPTANCE CRITERIA**

- [x]  by clicking the league button on the "create-event" page a form opens where I can configure the league I want to create
- [x]  In the form I can enter the following things
  - [x] Name (mandatory)
  - [x] Beschreibung
  - [x] choose from league or tournament (mandatory) is pre-selected based on which button was pressed in create event page
  - [x] Sportart (mandatory)
  - [x] Teams oder Einzelsportler -> if nothing is entered both are allowed
  - [x] Maximale Anzahl der teilnehmenden Teams -> if nothing is entered there is no limit
  - [x] Spielsystem (mandatory)
     - [x] if league: "Jeder-gegen-Jeden", "Hin- und Rückrunde " , "Schweizersystem" , "Dänisches System "
     - [x] if tournament: "K.-o. System" , "K.-o. System mit Gruppenphase ", "Double Elimination"
- [x] the form has two buttons
  - [x] create : this creates a league with the given parameters
  - [x] cancel : this cancels the process and brings you back to the create event page
- [x] by creating an league the user is remarked as organizer of this event
- [x] every event gets an unique id(mandatory, automatically assigned)
- [x] if any of the mandatory attributes is missing it is not possible to create the event and the user is prompted to enter them

GittiHab:
Why does the title of this story mention leagues _and_ tournaments but only describes later?

MarcelSchmidberger:
The creation process of leagues and tournaments only differs in a few points and therefore the two forms are basically the same. Therefore it was planned to build both forms for the basic event models in one go. This was intended to enable a different team to work on the tournament stuff later on, without messing with the creation process.

MarcelSchmidberger:
correct Problems stated in #232

tobiasduerschmid:
Please add the actual size the user story was worth.

usame-algan:
Actual size: M

PR #61, opened 2017-11-15
Bruteforce/events model

laurenzse:
League and Tournament are subclasses of Event. I couldn't find a way to validate that Event is not subclassed, i.e. an 'abstract class'. Maybe someone with more Ruby experiences has an idea?


app/models/event.rb
```diff
@@ -0,0 +1,4 @@
+class Event < ApplicationRecord
+    enum player_types: [:single, :team]
```
Askir: Is this actually neseccary or is Event going to have a `has_many: ;teams` association later anyways? Because if that is the case we could just implement `player_types = :single` by having a teamsize of 1.

db/migrate/20171114232449_remove_game_mode_from_tournament.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveGameModeFromTournament < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :tournaments, :game_mode
+    remove_column :tournaments, :game_mode, :integer
```
Askir: Okay,

- firstly: you don't need to make migrations for these changes if it's still in your branch. Nobody has created the database yet with your old migrations so you can just redo your initial `create_tournaments` or `creates_events` migrations without messing anything up.

- Secondly: If you ever have to do something like this because we already deployed something and now it's necessary to change the columns. I'd suggest changing all the affected tables in one migration (instead of three) so that the database stays in a consistent state. With rails migrations it's even possible to take the old values and move them to the new columns so that existing data doesn't get lost.

db/migrate/20171114225311_create_tournaments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTournaments < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tournaments do |t|
+      t.integer :game_mode
```
Askir: I have actually never needed to use enums yet, but they seem a good idea. Based on [this article](https://hackhands.com/ruby-on-enums-queries-and-rails-4-1/) using a column like: `t.integer :game_mode, default: 0, null: false` seems like a good idea to avoid null checks further down the line.

spec/models/event_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+describe "Event model", type: :model do
+
+    it "should not validate when everything is specified correctly" do
+        valid = League.new(name: "Weitwurfmeisterschaften",
```
Askir: Isn't this a typical case of using FactoryBot instead of hardcoding the values? If you want a certain value to not be defined just set it to nil.

spec/models/league_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+describe "League model", type: :model do
+
+    it "should get created with correct data" do
+        league = League.new(name: "Weitwurfmeisterschaften",
```
Askir: How about defining the object in a before/let block and then making different tests for all the expect statements? So that we know what field exactly is actually not working instead of having this one big test fail.

spec/models/tournament_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+describe "Tournament model", type: :model do
+
+    it "should get created with correct data" do
+        tournament = Tournament.new(name: "Weitwurfmeisterschaften",
```
Askir: Same as with leagues.

Askir (changes requested):

Askir:
@laurenzse 
regarding abstract classes: Ruby doesn't have these but you can somewhat imitate their behavior by raising a NotImplementedError on member methods. I'm also quite sure that you don't actually need a table for each of the subclasses. But for further info on that you should look into Single table inheritance.

erksch:
https://devblast.com/b/single-table-inheritance-with-rails-4-part-1

This is a short tutorial about single table inheritance :)

app/models/league.rb
```diff
@@ -0,0 +1,5 @@
+class League < Event
+    enum game_modes: [:round_robin, :two_halfs, :swiss, :danish]
```
derari: What is the purpose of separate League and Tournament *classes* when the only difference is which game modes are allowed?

spec/models/league_spec.rb
chrisma: This test mostly seems to check whether the constructor behaves as expected. The factory for league models that is also part of this PR isn't used. Maybe https://github.com/hpi-swt2/sport-portal/blob/dev/spec/models/match_spec.rb could be an inspiration?

app/models/event.rb
laurenzse: I like the idea! However, our PO wanted us to implement storing the "team size" right during event-creation. Right now, we don't really know when the team-assignment will happen during the process. Therefore, I would leave it as-is for now, and concentrate on getting a working solution until we have further details.

app/models/league.rb
laurenzse: From our POs explanations, we concluded that League and Tournament will have different behavior later on. Currently, you could even see the acceptance of different game modes as different behavior. Thus, I decided to create to separate classes (rather than managing the acceptance inside one class).

Is there any other reason, why we should keep those classes together?

laurenzse:
The league- and tournament-factory now have code-duplication with the event-factory. I couldn't figure out, how to reuse the event creation code within other factories. Any ideas?

spec/models/event_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+describe "Event model", type: :model do
+
+    it "should not validate without name" do
+        league = FactoryBot.build(:league)
+        league.name = nil
+
```
Askir: you can do this: `league = FactoryBot,build(:league, name: nil)` if you want to specifically test if the constructor works.

spec/factories/events.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+    factory :event do
+        sequence(:name) { |n| "name#{n}" }
+        sequence(:description) { |n| "description#{n}" }
+        sequence(:discipline) { |n| "discipline#{n}" }
+        player_type Event.player_types[Event.player_types.keys.sample]
+        max_teams { rand(1..30) }
```
Askir: Having a random number for this WILL lead to problems further down in testing abd make tests way more complex than they have to be. I'd put a static number of maybe 5 or 10.
Same with playertype.

Askir (approved):
Just 2 small things. Otherwise looks good to me.

Askir:
One more thing: please fix the 2 codeclimate issues ;)

spec/factories/events.rb
Askir: I just realized that we have random numbers in the matches factory aswell, out of my own experience with random data in tests I'm really not a fan of this. This can result in tests sometimes passing and sometimes failing depending on the random numbers and might be difficult to track down. So I'd still suggest removing these but I'll let you decide.

laurenzse: I would leave them, as our tests should ideally be able to cope with different numbers.

Askir: The problem with that is, that you have to essentially replicate the functionality that you are currently testing in your tests, instead of just hardcoding values. Which will make tests way more difficult to read and might also end up with you just copying the method you are currently testing. Which in the end defeats the purpose of testing at all.

tobiasduerschmid: Since I have been requested for feedback, I will add some general thought without knowing the context of your solution and the specific test cases. 
Random tests are indeed a bit different than usual f(input)=output tests. However, if you apply them with caution, they can be the most valuable tests you could write. Since they do not test for specific values but some out of a large range, they usually have a larger input coverage and, therefore, can find more defects, which is the ultimate goal of testing. However, the input range should, in the ideal case, equal all possible user-defined inputs for the subroutine. Than, each failing test is a defect and SHOULD be resolved. Hence, I do not agree with @Askir that sometimes passing and sometimes failing tests are bad. They are better than a test that is green since they identify misbehavior for some input. Sure, you have to output the chosen value to identify the bug. Otherwise the random value is useless. Random tests are especially useful if used as metamorphic tests (see testing slides or do individual research), when they define a relation between inputs and outputs instead of replicating the tested function. I did not look at your tests since no pointer was give to me. Hence I cannot say anything to your concrete implementation.
However, random tests with a small range of inputs (e.g., 1..30) might be transformed to a test that covers ALL of these inputs in one test, or in separate tests, depending on the kinds of inputs.

tobiasduerschmid (commented):
I added some thoughts about random testing without looking at your concrete tests.

Askir:
I did some basic merging but tests have to be fixed and the UI doesn't link to the correct model data anymore. @laurenzse please talk to @ConstantinLange about fixing all the errors you get right now. Also @usame-algan if you're done with the new creates event page consider merging it into this branch instead of into dev so we already use the new guidelines (one branch per user-story).

Askir:
@hpi-swt2/team-ms-17 one more review here please, since I now am a developer of this branch and reviewing your own code usually doesn't help anyone :)

app/controllers/events_controller.rb
```diff
@@ -53,6 +53,6 @@ def set_event
 
     # Only allow a trusted parameter "white list" through.
     def event_params
-      params.require(:event).permit(:name, :description, :gamemode, :sport, :teamsport, :playercount, :gamesystem, :deadline, :startdate, :enddate)
+      params.require(:event).permit(:name, :discipline, :description, :game_mode, :deadline, :startdate, :enddate).merge(player_type: Event.types.first)
```
laurenzse: What happened to `playercount`? Did you move it somewhere else or is it not necessary anymore?

app/views/events/_form.html.erb
```diff
@@ -22,28 +22,13 @@
...
-  </div>
-
-  <div class="field">
-    <%= form.label :playercount %>
-    <%= form.number_field :playercount, id: :event_playercount %>
-  </div>
-
-  <div class="field">
-    <%= form.label :gamesystem %>
-    <%= form.text_area :gamesystem, id: :event_gamesystem %>
```
laurenzse, line 46: Also, here :) Where are those input-areas now?

spec/factories/events.rb
```diff
@@ -4,6 +4,10 @@
         sequence(:description) { |n| "description#{n}" }
         sequence(:discipline) { |n| "discipline#{n}" }
         player_type Event.player_types[Event.player_types.keys.sample]
+        game_mode League.game_modes[League.game_modes.keys.sample]
```
laurenzse: Earlier, I wrote that Event should be somehow an abstract class, i.e. not valid when generated alone. Aren't we destroying this concept, when using the League game-modes to generate a valid Event class?

spec/models/event_spec.rb
```diff
@@ -1,26 +1,65 @@
...
 
     it "should not validate without game_mode" do
-        league = FactoryBot.build(:league, game_mode: nil)
-        league.game_mode = nil
+      league = FactoryBot.build(:league, game_mode: nil)
+      expect(league.valid?).to eq(false)
+    end
+
+    it "should have an attribute deadline" do
+      date = Date.new(2017,11,16)
```
laurenzse, line 22: Is it a good idea to use the hard-coded date from our factory here again? What happens if we change the factory?

Askir: This is merged code, basically copied from constantin I would've written the test differently anyways but I think it is fine for now.

app/controllers/events_controller.rb
Askir: playercount wasn't mentioned in the original issue so I just removed it. It looked like Constantin added it by himself. See issue 19: https://github.com/hpi-swt2/sport-portal/issues/19

app/views/events/_form.html.erb
Askir: Isn't game_mode and gamesystem the same thing?

app/controllers/events_controller.rb
laurenzse: Are you sure about this? 

> Maximale Anzahl der teilnehmenden Teams -> if nothing is entered there is no limit

I think it's just named differently in our model

spec/factories/events.rb
Askir: again this is also mostly so that constantins tests still run it's probably a good idea to move those tests to the league model though, I agree.

app/controllers/events_controller.rb
Askir: oh my bad, readding it

PR #80, opened 2017-11-17
Add default scope to only show active events on events index page.

### LINKED ISSUE #34: Only current events should appear on the "all events" page
**AS AN** user
**I WANT TO** see only events where the deadline for signing up has not passed yet when viewing the *all events* page
**SO THAT** I don't get confused about which events I can sign up for

## ACCEPTANCE CRITERIA
- [x] when opening the *all events* page only events where it is still possible to sign up should be displayed by default

sumpfranze:
fixes #34


ConstantinLange (approved):
Well done!

Issue #64, opened 2017-11-15
Add deadline input field to create view


sumpfranze:
Is part of User Story #53

- [ ] can be entered by entering the numbers
- [ ] can be entered by date picker (calendar view)

Issue #63, opened 2017-11-15
Add deadline field to event model


sumpfranze:
Is part of User Story #53

PR #82, opened 2017-11-17
Feature/issuenumber5/53

Jeilef:
Includes Tasks #63  und #64


spec/views/events/new.html.erb_spec.rb
```diff
@@ -31,6 +32,9 @@
       assert_select "input[name=?]", "event[playercount]"
 
       assert_select "textarea[name=?]", "event[gamesystem]"
+
+      assert_select "input[name=?]", "event[deadline]"
     end
   end
+
```
sumpfranze, line 39: best change ever.
delete plz :see_no_evil:

sumpfranze (approved):
LGTM :1st_place_medal:

PR #84, opened 2017-11-17
Update dependencies to remove vulnerability (rest-client 1.6.7)

fpottbaecker:
The issue was that even the most recent version of `coveralls` depends on `simple-cov ~> 0.14.1` and we were using `0.15.x`. The requirement was tightened from `simplecov >= 0.7` in [`coveralls 0.7.3`](https://github.com/lemurheavy/coveralls-ruby/blob/v0.7.3/coveralls-ruby.gemspec).

The issue was resolved by downgrading `simplecov` to `~> 0.14.0`


chrisma:
@tobiasduerschmid I remember talking about this security issue earlier, when it was detected by CodeClimate. Was the issue with updating that a newer version of `rest-client` was not compatible with the version of the `airbrake` gem that is required for error reporting to Errbit?

fpottbaecker:
@chrisma that depends on the version of `coveralls` at the time. 0.8.21 does not depend on `rest-client`.

Edit: `coveralls` did not require `rest-client` since `0.8.11`, which released 11.2015.

The issue was that even the most recent version of `coveralls` depends on `simple-cov ~> 0.14.1` and we were using `0.15.x`. The requirement was tightened from `simplecov >= 0.7` in `coveralls 0.7.3`.

chrisma (approved):
@Pottiman Thanks for the details!

fpottbaecker:
Considering this is a vulnerability, should we also apply this to `master`?

tobiasduerschmid:
It seems like the issue could be resolved without adapting the airbrake version. Therefore, it might be ok. 
@Pottiman: Remember to check whether errbit still works after the merge by reproducing or injecting a run time error on heroku

tobiasduerschmid (approved):
It seems like the issue could be resolved without adapting the airbrake version. Therefore, it might be ok.
@Pottiman: Remember to check whether errbit still works after the merge by reproducing or injecting a run time error on heroku

PR #92, opened 2017-11-20
33/issuenumber5/calendarsettings

ConstantinLange:
You can enter a startdate and an enddate for every event, as well as a duration. The validations check, that an enddate can not be before a startdate and the duration corresponds with the dates.


spec/models/event_spec.rb
```diff
@@ -1,14 +1,52 @@
...
+    expect(@event.duration).to eq(@duration)
   end
-  it "should have a attribute deadline" do
-  	@date = Date.new(2017,11,16)
-  	expect(@event.deadline).to eq(@date)
+
+  it "should have a deadline, startdate, enddate and duration" do
+    @event = Event.create(name: "NAME", description: "DESCRIPTION", gamemode: "GAMEMODE",
+    sport: "SPORT", teamsport: false, playercount: 1, gamesystem: "GAMESYSTEM",
+    deadline: nil, startdate: nil, enddate: nil, duration: nil)
```
sumpfranze: you can use the FactoryBot for that and just set the desired values to nil.

spec/models/event_spec.rb
```diff
@@ -1,14 +1,52 @@
...
 
 RSpec.describe Event, type: :model do
-  before(:context) do
-  	@event = FactoryBot.create :event
+
+it "should have an attribute deadline" do
+    @event = FactoryBot.create :event
+    @date = Date.new(2017,11,16)
+    expect(@event.deadline).to eq(@date)
+  end
```
sumpfranze: aren't "should have attribute" tests the tests for validation? So should also check, that the attribute can't be nil.

spec/models/event_spec.rb
sumpfranze: those four attributes should be tested for nil separately, since each validation for the attribute must be tested separately. This test would also be green, even if there is just a validation for one of the values.

Also it would give you more information on what exactly fails when you run the test.

spec/models/event_spec.rb
```diff
@@ -1,14 +1,52 @@
...
+  
+    expect(@event).not_to be_valid
+  end  
+
+  it "should be persistent that duration is the difference between start- and enddate" do
+    @event = Event.create(name: "NAME", description: "DESCRIPTION", gamemode: "GAMEMODE",
+    sport: "SPORT", teamsport: false, playercount: 1, gamesystem: "GAMESYSTEM",
+    deadline: Date.new(2017,11,16), startdate: Date.new(2017,12,01), enddate: Date.new(2017,12,05), duration: 6)
+
+    expect(@event).not_to be_valid
```
sumpfranze: you also need to test the positive. So create an event with a startdate, enddate and duration where the values are correct and test whether that is valid or not.

spec/models/event_spec.rb
```diff
@@ -1,14 +1,52 @@
...
   end
-  after(:context) do
-  	@event.destroy
+
+  it "should not be possible to have an enddate, that is before the startdate" do
+    @event = Event.create(name: "NAME", description: "DESCRIPTION", gamemode: "GAMEMODE",
+    sport: "SPORT", teamsport: false, playercount: 1, gamesystem: "GAMESYSTEM",
+    deadline: Date.new(2017,11,16), startdate: Date.new(2017,12,05), enddate: Date.new(2017,12,01), duration: 10)
+  
+    expect(@event).not_to be_valid
```
sumpfranze: It should be possible the other way around, so that there can be a valid event, when startdate < enddate.

sumpfranze (changes requested):
A general question, do we even need to save the duration in the database? Wouldn't it make way more sense, if the duration is just a value that is calculated in a function in the model? Would save a bit of headache with validations and reduce redundancy.

sumpfranze (approved):

PR #98, opened 2017-11-20
47/bruteforce/add create event button

### LINKED ISSUE #47: Add "create Event" button and selection page
Part of user story #18 

A button "create Event" should be added to the menu bar.
It should lead to a page where the user can choose between the two basic event categories "league" and "tournament"

CoderWombat:
closes #47


app/views/application/_navbar.html.erb
```diff
@@ -7,6 +7,8 @@
     <%= menu_item Match.model_name.human(:count => :many).titleize, matches_path %>
     <%= menu_item Event.model_name.human(:count => :many).titleize, events_path %>
     <%= menu_divider %>
+        <%= menu_item "Create Event", createEvent_path %>
```
Askir: Please use I18N standards for internationlization and don't hardcode strings. @erksch can you give a short tutorial maybe?

app/views/application/createEvent.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Choose event format</h1>
```
Askir: See above.

app/views/application/createEvent.html.erb
```diff
@@ -0,0 +1,4 @@
...
+
+<%= link_to '<button type="button">New league</button>'.html_safe, new_event_path %>
```
Askir: This should also definitely not be in dev untill new_event_paths is defined. So we should probably merge #19 with this before merging into dev. Otherwise this link will not be defined.

Askir (changes requested):

app/views/application/_navbar.html.erb
erksch: Please write your translations in a config/locales/de.yml like
```
de: 
  events:
    create_link: Veranstaltung erstellen 
```
If it does not exist, please create it.
In your view, you can access the translation with:
``<%= t 'events.create_link' %>``
or in your case:
``<%= menu_item t('events.create_link'), createEvent_path %>``

see https://phraseapp.com/blog/posts/rails-i18n-guide/ for more examples :)

config/routes.rb
```diff
@@ -19,4 +19,7 @@
   resources :users do
     get 'dashboard', on: :member
   end
+
+  #Define route for Create Event Button
+  get "/createEvent" , to: "application#createEvent" , as: "createEvent"
```
erksch: Please use snake case consistently for methods.
``..., as: "create_event"``
So you can redirect to the action with a ``create_event_path()`` method and not ``createEvent_path()``

app/views/application/_navbar.html.erb
CoderWombat: @erksch Is there a way to also use that method in
<%= link_to '<button type="button">New league</button>'.html_safe, new_event_path %>
to internationalize "New league" ?

Askir (approved):

Askir:
looks good now but maybe rebase the branch so it doesn't look so weird with +2k lines :D `git rebase dev`

Issue #56, opened 2017-11-14
Mail verification: unique email addresses


MerlindlH:
**As an** admin
**I want to** have only one user per e-mail address
**In order to** identify my users via e-mail addresses (for example during login processes)

**Acceptance Criteria**:
- [x] sign up with existing e-mail addresses is not possible
- [x] e-mail addresses are used all lowercase (M@a.de == m@a.de)

**Out of Scope**:
- student.hpi.uni-potsdam.de and student.hpi.de are not the same (maybe new userstory)

Markus-Brand:
Was already implemented

fpottbaecker:
Estimated size: 3 (Unit: 30 min)
Actual size: 2.42 (1.21h)

PR #106, opened 2017-11-21
#56/swteam/unique emails

Markus-Brand:
Branches from the cucumber branch
Files that are changed by this feature are:

features/step_definitions/user.rb
features/unique_email.feature
features/support/env.local.rb


features/step_definitions/content.rb
```diff
@@ -0,0 +1,8 @@
+Then(/^it should show the names of the participating teams$/) do
+  expect(page).to have_text(@matches.last.team_home.name)
```
DanielStephan97: 06_Testing.pdf slide 33: "Do not test for existence of actual text" - Does this mean you should rather use have_content than have_text?

features/unique_email.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Unique Emails
+In order to identify my users via e-mail addresses (for example during login processes)
```
DanielStephan97: Order is wrong?

features/unique_email.feature
```diff
@@ -0,0 +1,16 @@
...
+As an admin
+I want to have only one user per e-mail address
+
+  Scenario: Sign up with existing email address
+    Given a user a with email test@test.com
+    And a new user b with email test@test.com
+    When b tries to sign up
+    Then b should not be able to sign-up
```
DanielStephan97: Wording: maybe 'the sign up should fail'? It's also sign up, not sign-up.

features/step_definitions/content.rb
fpottbaecker: That actually referred to not testing text because it might be translated.

This tests whether the data from the model can be found on the page, which should be fine.

features/unique_email.feature
fpottbaecker: No, this is the default generated by features, user stories have another order.

features/unique_email.feature
fpottbaecker: @Markus-Brand says: This needs a reference to a user, so it needs to include a reference to the name b

And yes, it is, will fix.

features/unique_email.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Scenario: email addresses should be case insensitive
+    Given a user a with email TEST@TesT.cOm
+    And a new user b with email test@test.com
+    When b tries to sign up
+    Then b should not be able to sign-up
```
fpottbaecker: Please add newlines add the end of files.

features/unique_email.feature
ErikBrendel: please, don't start with the `In order to` - line. `As a *` should be the beginning of this description, for better readability

ErikBrendel (approved):

Issue #39, opened 2017-11-13
Profile Picture


PoJo93:
- #As logged in User 
- #IWantTo have a profile picture 
- #To
  - be quickly identified in player lists
  - to have a more personal touch
  - to get a personal touch
- #AcceptanceCriteria 
  - [ ] I can upload the profile picture on my profile page
    - The picture should not be larger than 2MB
    - A default picture should be shown which should be free to use when I haven't set a profile picture
  - [x] I want to see my profile picture next to my name when I am logged on on top right corner

nikriek:
Actual complexity of 40

Issue #27, opened 2017-11-11
Roles in team


svitlanakalancha:
Rights management in team. There are two roles in team: team owner and team member. 

Acceptance criteria:

- [x] Person who created the team gets a team ownership automatically

lheilmann:
@svitlanakalancha In contrast to our discussion in the last meeting, I decided to implement multiple m:n-Relations between User and Team for both roles 'TeamMember' **and** 'TeamOwner', because a team should be able to have multiple team members and multiple team owners, as well. So a team ownership is not stored in the team, it is stored in a separate join table. Same for a team membership.

lheilmann:
@svitlanakalancha At the moment the user creating the team gets a team ownership automatically (Criteria 2), only if the user is signed in. If not, no team ownership is created. Furthermore the team validation will fail, because a team must have minimum one owner (Criteria 1).

lheilmann:
@svitlanakalancha I assume, a team owner is also a team member. So while creating a new team, the current signed in user gets a team ownership **and** a team membership.

svitlanakalancha:
@lheilmann Regarding criteria 2, if the user isn't signed in, he can't add team at all. Can we implement such behavior?
The second assumption is correct :)

lheilmann:
@svitlanakalancha Yes, we can! That's why I have to talk to other teams, to clarify how route/controller method authorization should be implemented in general. It is definitely possible two restrict the user to create/update/destroy a team. Implementing this behaviour right now will cause many tests that were shipped with the application frame to fail. I will talk to the scrum masters to find a quick solution for that.

lheilmann:
@svitlanakalancha Criteria 3 collides technically with #25. I started to implement a first approach. But I will talk to assignees of #25 to determine intersection of the two issues.

bakoe:
@lheilmann Regarding the validations of `assign_membership_to` and `assign_ownership_to`: I think the methods provided by the already included [cancancan](https://github.com/CanCanCommunity/cancancan) gem could replace the quite “manual” checks introduced in 0c4face. 

Using cancancan, we could use something like:

```ruby
class Ability
  include CanCan::Ability

  def initialize(user)
    …
    if user.present?  # additional permissions for logged in users
      …
      can :assign_membership, Team, team_members: { member: { :id => user.id } }
      can :assign_ownership, Team, team_owners: { owner: { :id => user.id } }
      …
    end
  end
end
```

[...]

Issue #21, opened 2017-11-11
Create new team


svitlanakalancha:
As User I want to create a team in order to participate in sport events.

**Acceptance criteria:**

- [x] Team has following fields:
	-id(mandatory, automatically assigned)
	-name (mandatory) "Name"
	-description (optional) "Beschreibung"
	-kinds of sports (optional) "Sportarten"
	-private checkbox "Privat"(if private - visible only for team members, by default public - visible for all)

- [x] All fields are anytime editable by all team members.

- [x] Button "Save" ("Speichern") and "Cancel" ("Abbrechen"): visible only in edit modus

- [x] Button "Edit" ("Bearbeiten") : visible only in normal modus

- [x] all buttons are functionable

<img width="552" alt="screen shot 2017-11-11 at 21 09 05" src="https://user-images.githubusercontent.com/27727717/32693088-bf74015a-c724-11e7-8860-bdf2e92e7647.png">

bakoe:
Edit functionality depends on #27: *Roles in team*

PR #107, opened 2017-11-22
#21/kickerz/create new team

conradlempert:
Added some more features to creating new teams.

Things from the issue that are missing currently:
Adding team logo: #39 needs to be finished for that, so i can use the same implementation
Only team members can edit team: #27 needs to be finished for that, teams cant have team members yet.

The branch 21/kickerz/create-new-team (no # as first letter) is old and can be ignored.


app/controllers/teams_controller.rb
```diff
@@ -53,6 +53,6 @@ def set_team
 
     # Only allow a trusted parameter "white list" through.
     def team_params
-      params.require(:team).permit(:name)
+      params.require(:team).permit(:name, :private, :description, :kindofsport)
```
bakoe: Please use Snake case (`kind_of_sport`) throughout all variable and field names.

app/models/team.rb
```diff
@@ -4,10 +4,14 @@
 #
 #  id         :integer          not null, primary key
 #  name       :string
+#  description:text
```
bakoe: Please use a whitespace between the field names and types:

```ruby
description :text
```

bakoe: Please use a whitespace between the field names and types:

```ruby
description :text
```

app/models/team.rb
```diff
@@ -4,10 +4,14 @@
...
+#  kindofsport:string
+#  private    :boolean
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #
 
 class Team < ApplicationRecord
   validates :name, presence: true
+  validates :private, inclusion: { :in => [true, false] }
```
bakoe: According to a comment to https://stackoverflow.com/a/4721574, you can use the following shorter form instead:

```ruby
validates :private, inclusion: [true, false]
```

app/views/teams/_form.html.erb
```diff
@@ -10,6 +10,31 @@
...
+    <div class="col-lg-10">
+      <%= f.text_area :description, :class => 'form-control' %>
+      <%= f.error_span(:description) %>
+    </div>
+  </div>
+
+  <div class="form-group">
+    <div class="col-lg-offset-2 col-lg-10">
+      <label>
+        <%= f.check_box :private %>&nbsp;Private
```
bakoe: Please use i18n here instead of the hardcoded `Private`

app/models/team.rb
```diff
@@ -4,10 +4,14 @@
...
 #  name       :string
+#  description:text
+#  kindofsport:string
+#  private    :boolean
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #
 
 class Team < ApplicationRecord
   validates :name, presence: true
```
bakoe: Maybe the `private` field should be mandatory as well – to my mind, the checkbox is always either checked or not checked thus there should always be a value for the boolean `private` field when using the frontend view. By checking the `private` field's presence in the Team class, we could also force API users other than the frontend view to always fill the `private` field.

db/schema.rb
```diff
@@ -42,6 +42,9 @@
     t.string "name"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.text "description"
+    t.string "kindofsport"
+    t.boolean "private"
```
bakoe: See above: If you make the `private` field mandatory, it should have a `null: false` attribute in the db schema (as well as in the migration).

bakoe (changes requested):
In general I think it would be nice if there would be additional tests for the now extended teams form.

EDIT: @conradlempert informed me that the view fields are already filled automatically in the test. :+1:

bakoe (approved):

Issue #58, opened 2017-11-14
Open ID Integration: registration via OpenID


MerlindlH:
**As an** unregistered user with OpenIDAccount 
**I want to** sign up via OpenID
**In order to** simplify my login process

**Acceptance Criteria**:
- [x] after finishing the OpenID sign in you come to a page with an already partly filled form (depending on what openID transmits to our platform )

**Tasks**:
- [x] research
- [x] model adjustments
- [x] add signup via OpenID

fpottbaecker:
Merged in #109

fpottbaecker:
Estimated size: 13 (Unit: 30min)
Actual Size: 15.56 (7.78 h)

PR #109, opened 2017-11-22
#58/swteam/openid registration

fpottbaecker:
This will add the ability to sign in with an unlinked openID account, which prompts the user to fill in missing model data and creates an account once he submits the form.

This pull request also includes an initial german translation for the devise language file, as requested by the PO.


fpottbaecker:
This feature has been reviewed and accepted by the product owner, @MerlindlH.

DerIntergalaktischeGrosshorst (approved):
Looks fine

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -29,6 +28,17 @@ def sign_in_user(user)
       flash_if_navigational :notice, :success, provider: 'OpenID'
     end
 
+    def sign_up_user(auth)
+      session['omniauth.data'] = {
+        uid: auth.uid,
+        provider: auth.provider,
+        email: auth.info.email,
+        expires: Time.current + 10.minutes
```
ErikBrendel: why 10 minutes?

app/models/user.rb
```diff
@@ -47,7 +47,28 @@ def reset_omniauth
     self.provider = nil
   end
 
-  def self.from_omniauth(auth)
-    where(provider: auth.provider, uid: auth.uid).first
+  class << self
+    def new_with_session(_, session)
```
ErikBrendel, line 51: please add a comment: What is a session in this context

app/views/devise/registrations/new.html.erb
```diff
@@ -21,24 +21,26 @@
...
-  <div class="form-group">
-    <%= f.label :password_confirmation, :class => 'control-label col-lg-2' %>
-    <div class="col-lg-10">
-      <%= f.password_field :password_confirmation, class: 'form-control' %>
-      <%= f.error_span(:password_confirmation) %>
-    </div>
-  </div>
+  <% unless resource.has_omniauth? %>
+      <div class="form-group">
+        <%= f.label :password, :class => 'control-label col-lg-2' %>
```
ErikBrendel: :rocket: (also in line +7)

features/openid/registration.feature
```diff
@@ -0,0 +1,21 @@
+Feature: OpenID Registration
+  In order to simplify my login process
+  As an unregistered user with an OpenID Account
+  I want to sign up with my OpenID Account
+
+  Scenario: Sign up attempt
+    Given a new user
+    And an unlinked OpenID account with email 'test@test.com'
+    When the account is used to sign in
+    Then the 'E-Mail' input should be filled with 'test@test.com'
```
ErikBrendel: `should already be filled with` maybe? Should be filled sounds like an action (like 'the user should fill in now')

ErikBrendel (approved):

app/models/user.rb
fpottbaecker: This method is called by devise when a resource is built, which is done automatically by most controllers, or in this case the users_controller, with the call to `load_resource` call

PR #111, opened 2017-11-23
Flexibus/i18n german localization

jonashering:
Change default language to German and add German locales as yamls. Request of POs.

Locales are taken from:
* `devise.de.yml`: https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/de.yml
* `de.default.yml` : https://gist.github.com/thomasklemm/6821669
* `de.yml` and `de.helpers.yml` manually translated
(sources to links can also be found in files)

Note:
* Cucumber tests are currently set to English

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -12,6 +12,10 @@ module SportPortal
     config.load_defaults 5.1
 
   	config.generators.javascript_engine = :js
+
+    # Set default language to German
+    config.i18n.default_locale = :de
+
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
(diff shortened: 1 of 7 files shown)


jkbe:
@hpi-swt2/scrum-masters for your information.

jkbe (approved):
Looks good to me!

Askir:
Looks alright, although I'm not sure if we actually need the huge file with default translations it probably wont hurt. Also there are still a lot of spots where I18N isn't used yet but I hope that will be resolved in the next few weeks.

Askir (approved):

jkbe:
We already have big files with default English phrases which we don't all use - so it doesn't really make a difference I guess. However, we will discuss this in more detail in the meeting of Scrum Masters on Friday.
I saw these places as well where i18n isn't used yet. It seams like the displayed fields etc. are coming straight from the DB schema...

PR #113, opened 2017-11-24
37/dividebyzero/profile page

### LINKED ISSUE #37: Profile Page
- #As logged in User 
- #IWantTo configure my profile 
- #To
  - store and see my personal data
- #AcceptanceCriteria
 - [ ] When hovering on my profile name on the top right corner I get a dropdown menu which contains an item  "Abmelden",
When I click on "Abmelden" then I want to get logged out which substitutes the button sign out on the top right corner

manner:
Closes #37.

Feature name is misleading. This PR is mostly the dropdown menu in the navbar.


config/locales/de.yml
```diff
@@ -36,6 +36,9 @@ de:
     index:
       lead: "Wir bieten unglaubliche Lösungen für vielfältige Use Cases."
 
+  navbar:
+    menu: "Menu"
```
jkbe: Menü?

spec/application/navbar_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    it "has a dropdown menu" do
+      visit root_path
+      expect(page).to have_css(".navbar .dropdown")
+    end
+
+    it "dropdown menu shows menu button" do
+      visit root_path
+      expect(page).to have_css(".navbar .dropdown-toggle", text: "Menu")
```
jkbe: Better to use `t()` here I believe.

jkbe (changes requested):

jak-ing:
Closes #37.

jkbe:
In the User Story, it's implicated that the user name should be shown instead of 'Menu'. Has this requirement changed?

spec/application/navbar_spec.rb
jkbe: Move this file into spec/features/...?

config/locales/de.yml
manner: Pardon. You are correct

spec/application/navbar_spec.rb
manner: `t()` didn't work as expected in the specs, But this is not needed anymore, because the drop down menu shouldn't show `Menu`, but rather the first name of the user (as you pointed out correctly)

jkbe: Ok, I think prefixing `t` with the module name worked for me (`I18n.t(...)`).

jkbe (approved):

citron0xa9 (approved):

PR #117, opened 2017-11-24
55/swteam/forgot password

ErikBrendel:
With this PR, 'forgot-password' mails get sent by the devise-subsystem.
This works test-wise, and (thanks to the gmx-account we created) actually works in the real world. These code snippets (as seen in `config/environment/devise.rb`) might need to be adjusted post-merge.


app/views/devise/mailer/reset_password_instructions.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <p><%= t('.instruction', default: 'Someone has requested a link to change your password, and you can do this through the link below.') %></p>
 
-<p><%= link_to t('.action', default: 'Change my password'), edit_password_url(@resource, reset_password_token: @token, locale: I18n.locale) %></p>
+<p><%= link_to t('.action', default: 'Change my password'), edit_password_url(@resource, host: ENV['HOST_URL'], reset_password_token: @token, locale: I18n.locale) %></p>
```
fpottbaecker, line 5: You might want to set the host globally, as shown here: http://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views

features/forgot_password.feature
```diff
@@ -0,0 +1,16 @@
+Feature:
+  A working "Forgot-Password" mechanism
+
+  Scenario: A User starts the "Forgot-Password" Routine
+    Given a new user ulf with email ulf@ulf.de
```
fpottbaecker: You might want to create a step "a user [name] with email [email]" instead, so far the terminology has been user for an already stored user and new user for a user not in the database.

features/forgot_password.feature
```diff
@@ -0,0 +1,16 @@
...
+    And ulf is stored in the database
+
+    When ulf visits the password recovery page
```
fpottbaecker: Consider using a background for this scenario, since this seems like two scenarios...

features/step_definitions/user.rb
```diff
@@ -72,3 +72,51 @@
 Then(/^(.*) should not be able to sign up$/) do |username|
   expect(user_named(username).persisted?).to be_falsey
 end
+
+
+When(/^(.+) visits the password recovery page$/) do |username|
```
fpottbaecker: since this action does not require any named object as reference, may you should get rid of the parameter, maybe 'When he visits...'.

features/step_definitions/user.rb
```diff
@@ -72,3 +72,51 @@
...
+end
+
+And(/^submits$/) do
+  click_button I18n.t('devise.registrations.reset_password')
+end
+
+Then(/^(\w+) gets an Email with a recovery link$/) do |username|
+  user = user_named(username)
+  reset_password_mail = ActionMailer::Base.deliveries.last
+  expect(reset_password_mail.subject).to have_content('asswor') # (P|p)asswor(d|t)
```
fpottbaecker: Consider using the relevant I18n translation here instead

features/step_definitions/user.rb
```diff
@@ -72,3 +72,51 @@
...
+  expect(reset_password_mail.to[0]).to eq(user.email)
+  expect(reset_password_mail.body.to_s).to have_css('a')
```
fpottbaecker: Maybe check if the url of the link is correct.

features/step_definitions/user.rb
```diff
@@ -72,3 +72,51 @@
...
+end
+
+When(/^(\w+) clicks the recovery link$/) do |username|
+  reset_password_mail = ActionMailer::Base.deliveries.last
+  path_regex = /(?:"https?\:\/\/.*?)(\/.*?)(?:")/
+  path = reset_password_mail.body.match(path_regex)[1]
```
fpottbaecker: Try to use something like capybara for webpages here as well, one example might be this gem: https://github.com/DavyJonesLocker/capybara-email

features/support/env.local.rb
```diff
@@ -112,6 +112,11 @@ def sign_in(user)
     @current_user = user
   end
 
+  def sign_out
+    click_link I18n.t('devise.registrations.sign_out')
```
fpottbaecker: Maybe you could just visit the destroy session path instead...

config/environments/production.rb
```diff
@@ -1,3 +1,5 @@
+ENV['HOST_URL'] = 'sport-portal-dev.herokuapp.com'
+
 # Configure airbrake, error reporting to Errbit
```
fpottbaecker, line 3: What is the default option for the action mailer in production?

config/environments/development.rb
```diff
@@ -27,7 +29,21 @@
   end
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
+  # actually send emails
+  config.action_mailer.perform_deliveries = true
+
+  config.action_mailer.delivery_method = :smtp
```
fpottbaecker, line 36: For development purposes, could we use `:file` instead, or is that to complicated to test locally? http://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration

fpottbaecker (changes requested):

config/environments/development.rb
ErikBrendel: I'm against that. I was pretty happy to see real emails in my personal mail program even in development mode. With this setting, one can also locally verify that the smtp connection is set up correctly.

config/environments/production.rb
ErikBrendel: none for now. I will copy the dev settings and open an issue that another mail address should get acquired. What do you think?

features/forgot_password.feature
ErikBrendel: :+1:

features/forgot_password.feature
ErikBrendel: :-1:

features/step_definitions/user.rb
ErikBrendel: :+1:

features/step_definitions/user.rb
ErikBrendel: how?

app/views/devise/mailer/reset_password_instructions.html.erb
ErikBrendel: would prevent us from using localhost in dev env and onthoer host in production. With this approach, one can specify the url for each env which sounded better for me

features/step_definitions/user.rb
ErikBrendel: I tried and failed. Is this line of code so bad?

config/environments/production.rb
fpottbaecker: No, please revert the change below

features/step_definitions/user.rb
```diff
@@ -72,3 +72,48 @@
 Then(/^(.*) should not be able to sign up$/) do |username|
   expect(user_named(username).persisted?).to be_falsey
 end
+
+
```
fpottbaecker, line 87: Space — the final frontier. These are the voyages of the starship `develop`. Its ongoing mission — to program strange new apps — to seek out new smells and new optimizations — to boldly code where no man has coded before.

features/step_definitions/user.rb
fpottbaecker: Check for the token somewhere in the link?

app/views/devise/mailer/reset_password_instructions.html.erb
fpottbaecker: Ok, this can also be left for later

fpottbaecker (approved):

Issue #32, opened 2017-11-11
event page as organizer


MarcelSchmidberger:
**AS A** organiser of an event
**I WANT TO** be able to edit the settings of events I created when viewing their page
**SO THAT** I can change the settings of my events

**ACCEPTANCE CRITERIA**

- [x] organiser can visit the event page
- [x] on the top right there is an edit button
- [x] pressing the icon makes all fields editable
- [x] on the bottom of the page a *save* and *cancel* button appears if the organizer is in edit mode
- [x] if something was changed and the organiser saves the changes by pressing the *save* button the changes are saved and the organiser leaves the edit mode for the page
- [x] pressing the cancel button will drop any changes and also leave edit mode
- [x] only the organiser of an event is able to edit the respective event

PR #119, opened 2017-11-24
#32/in5/organizer event page

stlutz:
Prepare #32. Editing events still not restricted to organizers.


config/locales/de.yml
```diff
@@ -103,3 +102,4 @@ de:
       away: "Team 2"
       score: "Ergebnis"
       points: "Punkte"
+
```
Jeilef, line 105: such Change. Much wow

app/views/events/show.html.erb
```diff
@@ -17,8 +34,7 @@
 
 <%# app/helpers/button_helper.rb %>
 <%= back_btn %>
-
-<% if user_signed_in? %>
+  <% if user_signed_in? %>
```
Jeilef, line 37: use can can can

Jeilef (approved):
Feature implemented. Minor inconvieniences

PR #120, opened 2017-11-24
#43/flexibus/signup on login page

### LINKED ISSUE #43: Log-in Page
AS a user
I WANT to have possibility to create an account
SO THAT I can use sport portal

- Interface of sign-in page should look like on the mockup above (layout, texts, design doesn't matter)
- Link "Registrieren" redirect user to page "sign-up"

jonashering:
Closes #43.

Redesign login page as requested by our PO.


jkbe (changes requested):
German translation for 'remember me' is not shown although already available in gem `devise-bootstrap-views-0.0.11/locales/de.yml`

jkbe (approved):

Issue #29, opened 2017-11-11
Join event as team


svitlanakalancha:
**AS A** team owner
**I WANT TO** be able to join team events
**SO THAT** my team can participate in events

## ACCEPTANCE CRITERIA
Based on #85 
- [x] On the _all events_ page **and** detailed page of the event, if an event is a team sport there should be a button "join" as long as the user is not in a team participating at the event
- [x] If the user clicks join, an overlay should appear asking the user to select the team he would like to participate with
- [x] Only teams the user is an **owner** of and which are valid for this event (according to the options of the event) should be displayed
- [x] If the user is in no (valid) team there should be a message telling the user about this and offering to create a new team
- After the team has joined the event
  - [x] The button should change to "leave event"
  - [x] The "leave event" button should be greyed out (disabled) to all team members except the owners
  - [x] When the "leave event" button has been clicked and it was enabled the team should leave the event and it should not be labeled "Participating" anymore
  - [x] The team should have joined the event
  - [x] The event should be labeled with "Participating"
- [x] Every time the users open the _all events_ page all events, where a team they are in has joined, should be labeled with "Participating"
- [x] The "join" button is only visible as long as the deadline has not passed
- [x] Only logged in users can join events

bakoe:
depends on #49: *create Model for events*

bakoe:
depends on #27: *Roles in team*

GittiHab:
Mockups to this user story  
If the user wants to join a team event he needs to pick a team he wants to join with. Only teams he is owner of and which fulfill the event requirements are shown to him.
![join_as_team_1](https://user-images.githubusercontent.com/6453623/33240363-361ce326-d2b5-11e7-8132-f95f7845860d.jpg)
If the user is not an owner of any team, he gets a message explaining to him and offering to create an own team.
![join_as_team_2](https://user-images.githubusercontent.com/6453623/33240364-363d1416-d2b5-11e7-9f3d-a215464f78e2.jpg)

stlutz:
Todo list:
- [ ] fix error `ParameterMissing in EventsController#join`
- [ ] refactor routes
- [ ] change from users to single player teams
- [ ] test ui interactions
- [ ] rework cancancan authorizations

GittiHab:
There seems to be a bug with joining, see #270. Please try to have it working with this user story.

GittiHab:
There are **following issues**
- [ ] Bug #270 still exists
[...]

PR #121, opened 2017-11-24
85/issuenumber5/joineventassingleplayer

Jeilef:


app/controllers/events_controller.rb
```diff
@@ -45,14 +46,43 @@ def destroy
...
   end
 
+  # PATCH/PUT /events/1/join
+  def join
+    #@event = Event.find(params[:id])
+
+    @event.users << current_user
+
+    if @event.save
+      flash[:success] = "Your have successfully joined #{@event.name}!"
```
pawelsoundso: Typo in message: your ->you

app/controllers/events_controller.rb
```diff
@@ -32,6 +32,7 @@ def create
 
   # PATCH/PUT /events/1
   def update
+    print event_params
```
pawelsoundso: Shouldn't this be excluded, since it doesn't really serve any purpose except debugging?

app/views/events/index.html.erb
```diff
@@ -27,6 +27,16 @@
         <td>
           <%= link_to t('.show', :default => t("helpers.links.show")),
                       event_path(event), :class => 'btn btn-default btn-xs' %>
+          <% if (user_signed_in? && event.teamsport==false)%>
+              <% if event.users.where("user_id = ?", current_user.id).exists? %>
+                  <%= link_to 'Leave Event', join_event_path(event), :method => "put",
+                              :class => 'btn btn-default', id: :leavebutton%>
+                  Participating
```
pawelsoundso: Is localization now an impicit acceptance criteria or not? If yes we should maybe add localized button labels.

pawelsoundso (approved):
Functionality seems complete, localization and some debugging stuff could be deleted though.

app/controllers/events_controller.rb
sumpfranze: should be internationalized anyway.

config/routes.rb
```diff
@@ -1,6 +1,15 @@
...
 Rails.application.routes.draw do
-  resources :events
+
+  resources :events do
+    member do
+      put :join
+    end
+  end
+
+  # post '/events/:id/join', to: 'events#join'
```
sumpfranze: remove this + empty lines

db/migrate/20171124095557_create_join_table_events_users.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableEventsUsers < ActiveRecord::Migration[5.1]
+  def change
+    create_join_table :events, :users do |t|
+      # t.index [:event_id, :user_id]
+      # t.index [:user_id, :event_id]
```
sumpfranze, line 5: indices could be used for faster querying of data.

spec/features/event/index_event_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      sign_in @user
+      @singleevent = FactoryBot.create :event
+      @singleevent.deadline = Date.yesterday
+      visit events_path
+
+      expect(page).not_to have_button("Join Event")
+    end
+
+    it "should be not possible to join an event if the user is not logged in" do
+      @singleevent = FactoryBot.create :event
```
sumpfranze: is there no sign_out method in devise? if so, you could move the sign_in @user statements into the before block to reduce redundancy

spec/features/event/show_event_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  it "should be possible for a user to see, which event he/she joined" do
+    sign_in @user
+    visit event_path(@singleevent)
+    click_link("Join Event")
+
+    expect(page).to have_content("Participating")
+  end
+
+  it "should be not possible to join an event if the user is not logged in" do
+    visit event_path(@singleevent)
```
sumpfranze, line 41: same as in event/index

app/models/event.rb
```diff
@@ -21,6 +21,8 @@ class Event < ApplicationRecord
 
   scope :active, -> { where('deadline >= ?', Date.current) }
 
+  has_and_belongs_to_many :users
```
sumpfranze, line 37: pretty sure you can name an association with 
has_and_belongs_to_many :participants, source => :users

sumpfranze:
Discussion about this was, to create a single player team for single player events, when joining such an event, to not have to make a distinction of users and teams in the association. Could unify some code, especially when joining events with a team is implemented. Ref #29 

Would change this pr quite a bit

db/migrate/20171124095557_create_join_table_events_users.rb
Jeilef: I dont think that its necessary

app/models/event.rb
Jeilef: What else does it change in the Code?

spec/features/event/show_event_spec.rb
Jeilef: not exactly. this checks whether everything works on the Show page. The other test checks whether everything is fine on the index page

spec/features/event/index_event_spec.rb
Jeilef: there isnt a sign out method yet

app/views/events/index.html.erb
Jeilef: technically it is...

Jeilef:
Because the referenced Issue doenst Show any Progress yet, I would sugest that we merge this issue. Missing translations will be added later, but I want to have this big, possible conflicting chunk merged asap

app/models/event.rb
sumpfranze: Anywhere where you're using event.users to get the participants would change to event.participants

spec/features/event/show_event_spec.rb
sumpfranze: I know, meant the same as in the same comment about the sign out method. Sorry if it was unclear

spec/controllers/events_controller_spec.rb
```diff
@@ -25,15 +25,36 @@
...
+    sign_in @user
+  end
+
   # This should return the minimal set of attributes required to create a valid
   # Event. As you add validations to Event, be sure to
   # adjust the attributes here as well.
   let(:valid_attributes) {
-    skip("Add a hash of attributes valid for your model")
+    {
+        name: "TT Event",
```
Askir, line 38: Since I've seen this multiple times by now: please use `FactoryBot.build(:event).attributes` to create these hashes. Or if you want to make them specific for leagues or tournaments use `FactoryBot.build(:league/:event)` respectively. Makes it way easier to change the values in a single place if you do end up changing the model.

spec/models/event_spec.rb
```diff
@@ -45,21 +45,69 @@
...
+      @old_event = FactoryBot.create(:event, deadline: Date.yesterday)
 
       expect(Event.active).to include(@new_event)
-      expect(Event.active).to_not include(@old_event)
+      expect(Event.active).not_to include(@old_event)
       expect(Event.all).to include(@new_event, @old_event)
     end
 
+it "should have an attribute deadline" do
+    @date = Date.tomorrow
```
Askir, line 66: If you don't plan on using variables across tests there is no need to for making them instance variables. Just delete the @ :)

spec/views/events/show.html.erb_spec.rb
```diff
@@ -5,17 +5,35 @@
...
+    # expect(rendered).to match(/MyText/)
+    # expect(rendered).to match(/Gamemode/)
+    # expect(rendered).to match(/Sport/)
+    # expect(rendered).to match(/false/)
+    # expect(rendered).to match(/2/)
+    # expect(rendered).to match(/Gamesystem/)
+    # expect(rendered).to match(Date.tomorrow.to_s)
+    # expect(rendered).to match((Date.tomorrow+1).to_s)
+    # expect(rendered).to match((Date.tomorrow+3).to_s)
+    expect(true)
```
Askir, line 37: pretty sure that this test just doesn't do anything?

Askir (commented):
Some additional info on making the code a little cleaner. I also agree that this should be implemented differently but we will talk about that seperately.

Issue #28, opened 2017-11-11
Visibility in team


svitlanakalancha:
As User I can see the information about other teams.

Acceptance criteria:

- [x] all teams are by default public

- [x] nobody can see info about team except team members, event owner and admin, if team is private

- [x] private teams are not visible on the page with all teams 

- [x] pages of private teams are not accessible for not team members

bakoe:
depends on #27: *Roles in team*

bakoe:
Distinction between private and public teams depends on #21: *Create new team*

bakoe:
# Current status

## ☑️ Done:
- Added tests for the read abilities of logged-out and logged-in users (depending on whether they are team members/owners or not)
- Extended `ability.rb` to fulfill the requirements regarding the team reading permissions of logged-out and logged-in users
- Did successful manual acceptance tests to verify the abilities' influence on the teams' index and show views (logged-out, logged-in and logged-in as an owner/member of a private team)

## 📋 To Do:
- [ ] ~~Extend the abilities test for the read abilities of admin users (see #46)~~
- [ ] ~~Handle user's admin status in the `ability.rb` _model_ file (see #46)~~
- [x] Either comment in or delete the `create` and `update` abilities for logged-in users (see #21)
- [x] Either comment in or delete the `destroy` ability for logged-in users (see #22)
- [x] Either comment in or delete the `assign_ownership_to` and `assign_membership_to` abilities for logged-in users (see #24, #25)
- [x] **Make sure the existing tests for the teams controller don't fail anymore** (see #27)

## ⚠️ Needs clarification:
- Do we already make sure that teams “are by default public” (maybe via #21)? If not, how and where do we want to implement this functionality?
- Which issue/story is responsible for adding abilities for admin users? (it is already included in here: https://github.com/hpi-swt2/sport-portal/issues/46)

bakoe:
Regarding the first acceptance criteria (“all teams are by default public”): #28 made sure that all newly created teams are by default public. To ensure that already existing teams without a private/public property are also public, we could think about adding a database migration that assigns every team that is neither public nor private the “private: false“ property.

bakoe:
Regarding the second acceptance criteria (“nobody can see info about team except team members, event owner and **admin**, if team is private”): #46 introduces the admin role. We’ll try to make sure that #46 gives the admin(s) the ability to read (as well as create, update and destroy) every team, regardless of whether it is public or private.

birneamstiel:
[...]

Issue #25, opened 2017-11-11
Assign/delete team ownership


svitlanakalancha:
As Team Owner I can assign or delete team owner rights to other team members.

**Acceptance criteria:**

- [x] Only Team Owner can assign rights to other team members

- [x] Only Team Owner can delete rights of other team members

- [x] Team Owner can delete his own rights, if he is not the only one Team Owner

- [x]  Assign/delete rights button is visible only for Team Owners, buttons switch each other (if user is team owner -> delete rights, if not  -> assign rights)

bakoe:
depends on #27: *Roles in team*

lheilmann:
@jonasnoki @tomkellog Boys let's have a chat before you start implementing this feature! I started implementing a first approach of assigning a team ownership/membership to a user, because a criteria in #27 is colliding with some criterias of this issue. See progress in #27.

bakoe:
@jonasnoki @tomkellog Please also see [my comment in #27](https://github.com/hpi-swt2/sport-portal/issues/27#issuecomment-345540842) regarding the usage of the [cancancan gem](https://github.com/CanCanCommunity/cancancan) for defining abilities – could be useful for this user story.

niconomaa:
actual time: 15

niconomaa:
#### technical requirements for this issue

> Only Team Owner can assign/ delete rights to other team members
- check in `Ability` whether acting user is team's owner
> Team Owner can delete his own rights, if he is not the only one Team Owner
- check in `Ability` whether actions user is only remaining team owner when trying to perform action on himself
>  Assign/delete rights button is visible only for Team Owners, buttons switch each other (if user is team owner -> delete rights, if not -> assign rights)
- perform checks for all actions in view

Issue #23, opened 2017-11-11
Team page: information about team members


svitlanakalancha:
As Team member I see information about other team members on the team page.

**Acceptance criteria:** 

- [x] Team page includes a table with following columns:
        -number
	-name of the person
	-panel with actions (send email, delete, assign team owner)

- [x] All new team members are automatically shown in the table

- [x] User can choose group of users with a checkbox and perform action on them (delete, email)
	
- [x] Checkboxes to choose several team members

**Out of scope:** 
Implementation of delete and email functions

<img width="372" alt="screen shot 2017-11-11 at 21 41 47" src="https://user-images.githubusercontent.com/27727717/32693310-2905d3f6-c729-11e7-9726-95c34593b688.png">

bakoe:
depends on #27: *Roles in team*

niconomaa:
### technical requirements of this issue:

> Team page includes a table with [...] columns:

- iterate over each team member
- insert action item and properties for each target member
- disable buttons depending on user's `Ability` to perform action on target member
    - check each action's `Ability` using CanCanCan`s `can?` method

> All new team members are automatically shown in the table

- load members upon sending `show` action of `TeamsController`

> User can choose group of users with a checkbox and perform action on them (delete, email) 
#### and
>  Checkboxes to choose several team members
- insert checkbox while iterating over members
- insert additional checkbox brought to youuuu by jQuery to (de)select all members at a time
- introduce new action in `TeamsController` to perform existing action functionality on the selected set of target members
   - extract existing actions' functionalities to be shared among perform on single and multiple 
   - rewrite `Ability` to make it work with block parameters only
   - explicitly reload team for each target member to avoid automatic model caching
   - catch members on which action could not be performed due to missing `Ability`
   - display target members on whom action could be performed to user

- write tests for all of this

PR #130, opened 2017-11-26
24/kickerz/delete team member

niconomaa:
deletion functionality for owners and members; includes current `dev`


app/models/ability.rb
```diff
@@ -34,8 +34,41 @@ def initialize(user)
...
+
+  def num_owners(team, team_member)
+    if team.owners.include? User.find(team_member)
+      owners_after_delete = team.owners - [(User.find(team_member))]
+    else  
+      owners_after_delete = team.owners
+    end
+    return owners_after_delete.length
+  end
+
```
lheilmann: Just an idea to refactor this method:
```
def number_of_owners_after_delete(team, team_member_id)
    if team.owners.include? User.find(team_member_id)
      return team.owners.length - 1
    end
    return team.owners.length
end
```

Gemfile.lock
```diff
@@ -61,7 +61,7 @@ GEM
       railties (>= 3.0)
     builder (3.2.3)
     byebug (9.1.0)
-    cancancan (2.1.2)
+    cancancan (2.1.1)
```
lheilmann, line 64: Is there a reason why `cancancan` is downgraded?

app/models/ability.rb
```diff
@@ -35,7 +35,43 @@ def initialize(user)
...
+      owners = team.owners
+      another_user = User.find(team_member)
+      if owners.include? another_user
+        owners_after_delete = owners - [another_user]
+      else
+        owners_after_delete = owners
+      end
+      owners_after_delete.length
+    end
+
```
lheilmann: Consider refactoring this method:
```
  def self.number_of_owners_after_delete(team, team_member_id)
      current_length = team.owners.length
      team_member = User.find(team_member_id)

      if owners.include? team_member
        return current_length - 1
      end

      return current_length
    end
```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
bakoe, line 1: Please add this VS Code specific file to the `.gitignore` and thus remove it from this PR.

Gemfile.lock
niconomaa: absolutely not

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
     redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
   end
 
+  # Assigns team ownership to a specific team member
+  def assign_ownership
+    set_team
+    authorize! :assign_ownership, @team
+
+    # Checks whether the specified team member already has team ownership
+    if !@team.owners.include? (User.find(params[:team_member]))
```
bakoe: I think an `unless` instead of `if ! …` is better understandable.

bakoe: I think that an `unless` would be better readable than an `if !…` at this point. However, that's just my personal opinion.

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
-
     @team.destroy
-
     redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
   end
 
+  # Assigns team ownership to a specific team member
+  def assign_ownership
+    set_team
+    authorize! :assign_ownership, @team
```
bakoe: Because you explicitly defined the equally-named `:assign_ownership` ability, you could also use one central `load_and_authorize_resource :team` in the beginning of the controller instead of authorizing the team resource in each method separately. However, if you do so, you should test whether this behaves exactly in the same way as your current approach (which is also totally fine).

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
     TeamMember.where(team_id: team_id).delete_all
-
     @team.destroy
-
     redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
   end
 
+  # Assigns team ownership to a specific team member
+  def assign_ownership
+    set_team
```
bakoe: You can avoid this line (and all the other `set_team` lines in this file) by adding this method to the ones that have `before_action :set_team` assigned (see line 2 of this file):

```ruby
before_action :set_team, only: […, :assign_ownership]
```

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+
+    # Checks whether the specified team member already has team ownership
+    if !@team.owners.include? (User.find(params[:team_member]))
+      @team.owners << User.find(params[:team_member])
+      redirect_to @team
+    end
+  end
+
+  def delete_ownership
+    set_team
```
bakoe: You can avoid this line (and all the other `set_team` lines in this file) by adding this method to the ones that have `before_action :set_team` assigned (see line 2 of this file):

```ruby
before_action :set_team, only: […, :delete_ownership]
```

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+    authorize! :delete_ownership, @team
```
bakoe: Because you explicitly defined the equally-named `:delete_ownership` ability, you could also use one central `load_and_authorize_resource :team` in the beginning of the controller instead of authorizing the team resource in each method separately. However, if you do so, you should test whether this behaves exactly in the same way as your current approach (which is also totally fine).

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+    user = User.find(params[:team_member])
+
+    # Checks whether the specified team member does not have team ownership
```
bakoe: To my mind, this comment is both unnecessary and misleading: it says that the following line tests whether the team member does **not** have team ownership – what the line really does is testing whether the team member **has** a team ownership.

I think you can simply remove this comment. The following line itself is explanatory enough in my opinion. :+1:

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
-
     redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
   end
 
+  # Assigns team ownership to a specific team member
+  def assign_ownership
+    set_team
+    authorize! :assign_ownership, @team
+
+    # Checks whether the specified team member already has team ownership
```
bakoe: I'd prefer if this comment would
- either say what the condition for going into the if-branch is (e. g. “Checks whether the specified team member is not an owner yet”) or
- be removed (because the following line is quite self-explanatory).

In the current situation, the comment does describe what is necessary to **not** go into the if-branch, which is rather confusion in my opinion. I would prefer removing the comment.

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+
+    # Checks whether the specified team member does not have team ownership
+    if @team.owners.include?(user)
+      @team.owners.delete(user)
+      redirect_to @team
+    end
+  end
+
+  def delete_membership
+    set_team
```
bakoe: You can avoid this line (and all the other `set_team` lines in this file) by adding this method to the ones that have `before_action :set_team` assigned (see line 2 of this file):

```ruby
before_action :set_team, only: […, :delete_membership]
```

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+    authorize! :delete_membership, @team
```
bakoe: Because you explicitly defined the equally-named `:delete_membership` ability, you could also use one central `load_and_authorize_resource :team` in the beginning of the controller instead of authorizing the team resource in each method separately. However, if you do so, you should test whether this behaves exactly in the same way as your current approach (which is also totally fine).

app/controllers/teams_controller.rb
```diff
@@ -49,12 +49,51 @@ def destroy
...
+
+    if @team.owners.include? (user)
+      @team.owners.delete(user)
+    end
+
+    @team.members.delete(user)
+    redirect_to @team
+  end
+
+  def current_ability
```
bakoe, line 77: I don't see that this method is used anywhere explicitly. Is it used by some ruby/rails/cancancan magic in the background? If not, it could be removed I guess.

app/models/ability.rb
```diff
@@ -35,7 +35,42 @@ def initialize(user)
     #   end
 
     user ||= User.new # guest user (not logged in)
+    id = user.id
+
     # All users can only update their own user attributes
-    can :update, User, id: user.id
+    alias_action :create, :read, :update, :destroy, :to => :crud
+
+    can :crud, Team
```
bakoe: That's okay for the moment but will be replaced by the changes from #28 either way. Please keep that in mind when merging the `dev` (then probably containing the changes from #28) into you PR. :wink:

app/models/ability.rb
```diff
@@ -35,7 +35,42 @@ def initialize(user)
...
+      team.owners.include? user
+      team.owners.length > 1
+    end
+
+    can :delete_membership, Team, Team do |team|
+      team.owners.include? user and Ability.num_owners(team, team_member) > 0
+    end
+
+    can :delete_membership, Team, Team do |team|
+      Ability.num_owners(team, team_member) > 0 and Integer(id) == Integer(team_member)
```
bakoe: Please change the line into

```ruby
Integer(id) == Integer(team_member) and Ability.num_owners(team, team_member) > 0
```

(same logic, only swapped the two conditions) to improve comparability to line 56 when reading the code.

bakoe: Regarding `Integer(id)`: It seems to me that `id` is already an integer, isn't it? This would avoid at least one of the two conversions. The other conversion seems to be necessary because the POST parameter `team_member` is a string – right?

app/models/ability.rb
```diff
@@ -35,7 +35,42 @@ def initialize(user)
...
+    end
   end
+
+  private
+
+  def self.num_owners(team, team_member)
```
bakoe: It seems like the indentation of this method has gone wrong … to my mind, it should look like:

```ruby
  private
    def self.num_owners(team, team_member)
      owners = team.owners
      another_user = User.find(team_member)
      if owners.include? another_user
        owners_after_delete = owners - [another_user]
      else
        owners_after_delete = owners
      end
      owners_after_delete.length
    end
```

bakoe: Can you maybe find a more descriptive name for this method? Maybe something like `number_of_owners_after_deleting_member`?

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
...
+<table class="table table-striped">
+  <thead>
+    <tr>
+    </tr>
+  </thead>
+  <tbody>
+    <% @team.members.each do |member| %>
+      <tr>
+        <td>
+            <%= member.id%> <%= member.first_name%> <%= member.last_name%>
```
bakoe: This and the following lines are indented two spaces too far (file uses two-space indentation).

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
...
+            <%= link_to t('.delete_ownership', :default => t("helpers.links.delete_ownership")),
+            delete_ownership_team_path(@team, :team_member => member.id), :method => 'post', :class => 'btn btn-default btn-xs' if @team.owners.include? member %>
+
+            <%= link_to t('.assign_ownership', :default => t("helpers.links.assign_ownership")),
+            assign_ownership_team_path(@team, :team_member => member.id), :method => 'post', :class => 'btn btn-default btn-xs' if !@team.owners.include? member  %>
+
+
+
+
+            <%= link_to t('.delete_membership', :default => t("helpers.links.delete_membership")),
```
bakoe: Maybe consider using

- the bootstrap `btn-danger` class (to make the button appear red because it deletes something)
- a confirmation dialog as in
```ruby
<%= link_to t('.destroy', :default => t("helpers.links.destroy")),
            team_path(team),
            :method => :delete,
            :data => { :confirm => t('.confirm', :default => t("helpers.links.confirm")) },
            :class => 'btn btn-xs btn-danger' %>
``` 
(taken from `teams/index.html.erb`)

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
   <dd><%= @team.description %></dd>
 </dl>
 
+<table class="table table-striped">
+  <thead>
```
bakoe: I think you can remove the empty `<thead>` element.

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
...
+      <tr>
+        <td>
+            <%= member.id%> <%= member.first_name%> <%= member.last_name%>
+            
+            <%= link_to t('.delete_ownership', :default => t("helpers.links.delete_ownership")),
+            delete_ownership_team_path(@team, :team_member => member.id), :method => 'post', :class => 'btn btn-default btn-xs' if @team.owners.include? member %>
+
+            <%= link_to t('.assign_ownership', :default => t("helpers.links.assign_ownership")),
+            assign_ownership_team_path(@team, :team_member => member.id), :method => 'post', :class => 'btn btn-default btn-xs' if !@team.owners.include? member  %>
+
```
bakoe: Three empty lines that can be omitted …

config/locales/de.helpers.yml
```diff
@@ -18,6 +18,9 @@ de:
       new: "Neu"
       edit: "Bearbeiten"
       show: "Anzeigen"
+      assign_ownership: "Als Captain hinzufügen"
```
bakoe, line 21: Is the German word “Captain” for team owners common sense? Maybe also a question towards our PO @svitlanakalancha?

config/locales/de.helpers.yml
```diff
@@ -18,6 +18,9 @@ de:
...
+      delete_ownership: "Als Captain löschen"
+      delete_membership: "Löschen"
```
bakoe: Analogous to the English localization I would prefer something like “Aus dem Team entfernen” to clarify that not the user himself but only the membership is going to be deleted.

config/routes.rb
```diff
@@ -33,6 +33,24 @@
     get 'unlink', on: :member
   end
 
+  resources :teams do
```
bakoe: The three `resources :team do` blocks can be simplified to:

```ruby
  resources :teams do
    member do
      post :assign_ownership
      post :delete_ownership
      post :delete_membership
    end
  end
```

config/routes.rb
```diff
@@ -33,6 +33,24 @@
...
+    member do
+      post :assign_ownership
+    end
+  end
+
+  resources :teams do
```
bakoe: _See comment to line 36_

config/routes.rb
```diff
@@ -33,6 +33,24 @@
...
+    member do
+      post :delete_ownership
+    end 
+  end
+
+  resources :teams do
```
bakoe: _See comment to line 36_

spec/controllers/teams_controller_spec.rb
```diff
@@ -139,4 +139,84 @@
...
+    it 'fails when when not called as a team owner' do
+      team = Team.create! valid_attributes
+      anotherUser = FactoryBot.create :user
+
+      expect {
+        post :assign_ownership, params: { id: team.id, team_member: anotherUser.id }
+      }.to raise_error(CanCan::AccessDenied)
+    end
+
+    it 'should not add a team owner twice' do
```
bakoe, line 172: Nice! :+1:

spec/models/team_member_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+RSpec.describe TeamMember, type: :model do
+  it "should not be able to assign team ownership to users and delete it" do
+    team = FactoryBot.create :team
```
bakoe: You could factor out this and the following line (that is currently duplicated in every `it … do` block in this file) by using something like:

```ruby
RSpec.describe TeamMember, type: :model do
  describe 'when logged in' do
    let(:user){ FactoryBot.create :user }
    let(:team){ FactoryBot.create :team }

    it "should not be able to assign team ownership to users and delete it" do
      ability = Ability.new(user)
      assert ability.cannot?(:assign_ownership, team)
    end
    …
```

spec/models/team_member_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    user = FactoryBot.create :user
+    another_user = FactoryBot.create :user
+    team.members << user
+    team.members << another_user
+
+    ability = Ability.new(another_user, user.id)
+
+    assert ability.cannot?(:delete_membership, team)
+  end
+p
```
bakoe: Remove this `p` plz :wink:

spec/models/team_owner_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+RSpec.describe TeamOwner, type: :model do
+  it "should be able to assign team ownership to users and delete it" do
+    team = FactoryBot.create :team
```
bakoe: See the comment on the `team_member_spec.rb` file!

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
     expect(rendered).to have_content(@team.kind_of_sport, count: 1)
     expect(rendered).to have_content(@team.description, count: 1)
   end
+
+  it " shows assign ownership button for mere members" do
+    user = FactoryBot.create :user    
+    @team.members << user
+    render
+    rendered.should have_content('Als Captain hinzufügen')
```
bakoe: Please use internationalization-compatible string matching instead of the hardcoded `'Als Captain hinzufügen'`! There are some examples in the `dev`'s existing codebase I think.

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
...
+  end
+
+  it "does not show delete ownership button for mere members" do
+    user = FactoryBot.create :user    
+    @team.owners = []
+    render
+    rendered.should_not have_content('Als Captain löschen')
```
bakoe: _See comment to line 19_

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
...
+  end
+
+  it "shows delete ownership button for owners" do
+    user = FactoryBot.create :user    
+    @team.members << user
+    @team.owners << user
+    render
+    rendered.should have_content('Als Captain löschen')
```
bakoe: _See comment to line 19_

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
...
+  end
+
+  it "does not show assign ownership button for owners" do
+    user = FactoryBot.create :user    
+    @team.members << user
+    @team.owners << user
+    render
+    rendered.should_not have_content('Als Captain hinzufügen')
```
bakoe: _See comment to line 19_

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
     expect(rendered).to have_content(@team.kind_of_sport, count: 1)
     expect(rendered).to have_content(@team.description, count: 1)
   end
+
+  it " shows assign ownership button for mere members" do
```
bakoe: Leading whitespace in spec title can be removed

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
...
+    user = FactoryBot.create :user
```
bakoe: See the comment on the `team_member_spec.rb` file: you can factor out the redundant creation of the user via a `describe` and `let` construct.

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +11,34 @@
...
+    @team.members << user
+    @team.owners << user
+    render
+    rendered.should have_content('Als Captain löschen')
+  end
+
+  it "does not show assign ownership button for owners" do
+    user = FactoryBot.create :user    
+    @team.members << user
+    @team.owners << user
```
bakoe: If the factory for the `team` model would change and create a team with some members that are no team owners, this test would fail. Thus, I would replace this line by 

```ruby
@team.owners = @team.members
```

This would make sure that every member of the team is also an owner. By this means, there would definitely be no assign_ownership button.

bakoe (changes requested):

bakoe:
In general I'm afraid that part of the fifth acceptance criteria of #25 is not fully implemented yet:

> Assign/delete rights button is visible only for Team Owners, […]

At the moment, the assign/delete ownership buttons are also visible for team members that are no team owners. If you click one of these buttons as a mere user, you get an “CanCan::AccessDenied” error page.

EDIT: Maybe this is part of #23 instead of #25. Then, it should be clarified both issues by our PO @svitlanakalancha.

app/controllers/teams_controller.rb
niconomaa: it is assumed by CanCan and we are overwriting it here to pass the team_member param

app/models/ability.rb
niconomaa: yup -- just making sure here; but you're right; technically `id`'s fine as it is.

app/models/ability.rb
niconomaa: oh! i thought i had done this as originally proposed by @lheilmann. will rename.

app/controllers/teams_controller.rb
bakoe: Ah, alright. Nice find! :+1:

app/models/ability.rb
bakoe: **Update:** #28 was merged into `dev` some minutes ago – if you merge the `dev` into your feature branch for the next time, please make sure that you remove the following lines from your branch:

```ruby
alias_action :create, :read, :update, :destroy, :to => :crud
```

and

```ruby
can :crud, Team
```

config/locales/de.helpers.yml
niconomaa: I am pretty certain it isn't -- 'Besitzer', i thought, would've made it sound worse, though lol
let's talk about this tmrw

app/models/ability.rb
```diff
@@ -53,19 +53,18 @@ def initialize(user, team_member = nil)
     end
 
     can :delete_membership, Team, Team do |team|
-      team.owners.include? user and Ability.num_owners(team, team_member) > 0
+      team.owners.include? user and Ability.number_of_owners_after_delete(team, team_member) > 0
     end
 
     can :delete_membership, Team, Team do |team|
-      Ability.num_owners(team, team_member) > 0 and Integer(id) == Integer(team_member)
+      Integer(id) == team_member and Ability.number_of_owners_after_delete(team, team_member) > 0
```
bakoe: I think you accidentally removed the Integer conversion from `team_member` (where it ist necessary) and left it on `id` (where it can be omitted).

app/models/ability.rb
```diff
@@ -35,7 +35,41 @@ def initialize(user)
...
+      team.owners.include? user
+      team.owners.length > 1
+    end
+
+    can :delete_membership, Team, Team do |team|
+      team.owners.include? user and Ability.number_of_owners_after_delete(team, team_member) > 0
+    end
+
+    can :delete_membership, Team, Team do |team|
+      Integer(id) == team_member and Ability.number_of_owners_after_delete(team, team_member) > 0
```
bakoe: I think you accidentally removed the Integer conversion from `team_member` (where it ist necessary) and left it on `id` (where it can be omitted).

bakoe (changes requested):

.vscode/settings.json
bakoe: Somehow the (now ignored) file is still included in the PR – can you check again and make sure it is not included when merging the PR into the `dev`? @tomkellog

bakoe (approved):

app/models/ability.rb
```diff
@@ -35,17 +35,61 @@ def initialize(user)
...
+      can :assign_ownership, Team, Team do |team|
+        team.owners.include? user
+      end
+    end
+
+    def can_delete_membership(team_member, user)
+      can :delete_membership, Team, Team do |team|
+        user_id = user.id
+        exist_owners_after_delete = owners_after_delete = Ability.number_of_owners_after_delete(team, team_member) > 0
+        ((team.owners.include? user) && exist_owners_after_delete) || ((user_id == Integer(team_member)) && exist_owners_after_delete)
```
birneamstiel, line 75: - @tomkellog we had another fundamental talk right after you left, let's talk tomorrow or later :-)
- to summarize:
  - Let's try to reduce complexity of ability to only contain simple access abilities, like "can the current user access members of a specific team"
  - logic for allowing more complex actions (e.g. delete membership) will be added to the specific controller

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,31 @@
...
 </dl>
 
+<table class="table table-striped">
+  <tbody>
+    <% @team.members.each do |member| %>
+      <tr>
+        <td>
+          <%= member.id%> <%= member.first_name%> <%= member.last_name%>
+            
+          <% if can? :delete_ownership, @team %>
```
birneamstiel, line 28: - reduce "cleverness" of view by
  - do the "can user delete ownership?" check in controller
  - let controller prepare `@team.members` list to either contain an attribute (e.g. `deletable_ownership`) or have two lists: members and deletable_ownership_members (first option prefered)

PR #133, opened 2017-11-26
25/kickerz/assign and delete team ownership

niconomaa:
ownership assignment functionality, includes current `dev`


app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
...
+  </thead>
+  <tbody>
+    <% @team.members.each do |member| %>
+      <tr>
+        <td>
+            <%= member.id%> <%= member.first_name%> <%= member.last_name%>
+            
+            <% if @team.owners.include? member %>
+            <%= link_to t('.delete_ownership', :default => t("helpers.links.delete_ownership")),
+            delete_ownership_team_path(@team, :team_member => member.id), :method => 'post', :class => 'btn btn-default btn-xs' %>
```
lheilmann, line 34: Consider using a different naming like `team_member_id` or `member_id`

lheilmann: Did you define the required parameters for the `delete_ownership` route somewhere, e.g. to throw an exception when a invalid parameter type is given in a specific view?

app/views/teams/show.html.erb
```diff
@@ -18,6 +18,33 @@
...
+    <tr>
+    </tr>
+  </thead>
+  <tbody>
+    <% @team.members.each do |member| %>
+      <tr>
+        <td>
+            <%= member.id%> <%= member.first_name%> <%= member.last_name%>
+            
+            <% if @team.owners.include? member %>
```
lheilmann, line 32: Is it possible to check for the created ability `assign_ownership` in the view? See https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities.

spec/controllers/teams_controller_spec.rb
```diff
@@ -113,5 +121,4 @@
       expect(response).to redirect_to(teams_url)
     end
   end
-
```
lheilmann, line 116: Consider adding tests for the newly created controller actions/routes `assign_ownership` and `delete_ownership`. Have look at other tests in this file. An idea:

```
describe "POST #delete_ownership" do
    it "deletes a ownership of a team member" do
      team = FactoryBot.create :team
      user = FactoryBot.create :user
      another_user = FactoryBot.create :user

      team.owners << user
      team.owners << another_user
      team.members << user
      team.members << another_user

      expect {
        post :delete_ownership, params: { team_owner: another_user.id }
      }.to change(TeamOwner, :count).by(-1)
       .and change(TeamMember, :count).by(0)
    end

    it "redirects to the teams list" do
      team = FactoryBot.create :team
      user = FactoryBot.create :user
      another_user = FactoryBot.create :user

      team.owners << user
      team.owners << another_user
      team.members << user
      team.members << another_user

      post :delete_ownership, params: { team_owner: another_user.id }
      expect(response).to redirect_to(teams_url)
    end
  end
```

lheilmann: A problem I detected: Objects created by a factory do not have an ID.

config/routes.rb
```diff
@@ -27,5 +27,18 @@
     get 'unlink', on: :member
   end
 
+  resources :teams do
+    member do
+      post :assign_ownership
```
lheilmann, line 32: Consider using a different route definition, e.g. `post 'teams/assign_ownership/:member_id' => 'teams#assign_ownership'`

app/views/teams/show.html.erb
niconomaa: wrong params throw an exception #dontscrewmyroutes

app/views/teams/show.html.erb
niconomaa: you could -- but these authorizations logically overlap and are thus are semantically connected to these buttons. you actually _want_ to check, whether a team member is a user nor nawt.

app/views/teams/show.html.erb
niconomaa: should be stylize #hashtags in comments like _#so_ or **#so**? opinions please!

spec/controllers/teams_controller_spec.rb
niconomaa: oh fanks! didn't know that -- might help me a great deal for those f***ed up controller-tests with CanCan -- which, btw, are not in themselves best practice and are often recommended not to be written at all for there are cumbersome pieces of code. but whatever test coverage requirements dictate must.be.obeyed.

config/routes.rb
niconomaa: might wanna have a look at this together with me -- since this is IMO not (easily) possible. we are using GET parameters intentionally since the id in the url is used for the team and not for the member -- also GET parameters are easily accessed within Abilities.
will look into it, tho

niconomaa:
this is now fully included in https://github.com/hpi-swt2/sport-portal/pull/130 and can be revieved and merged together.

PR #150, opened 2017-11-27
27/kickerz/roles in team

lheilmann:


lheilmann:
Letzter offener Punkt: **Validierung** des Team-Models bez. der Existenz mindestens eines TeamOwners bzw. TeamMembers. Hierzu finden sich im Model die entsprechenden Zeilen. Diese Validierung führt allerdings dazu, dass viele Controller-Tests fehlschlagen. Die Team-Objekte in den Tests (`sport-portal/spec/controllers/teams_controller_spec.rb`) werden folgendermaßen generiert: `team = Team.create! valid_attributes`, wobei `valid_attributes` folgendermaßen generiert werden: `let(:valid_attributes) {
    FactoryBot.build(:team).attributes
  }`. Die Assoziationen für TeamOwner und TeamMember sind dabei nicht enthalten, was dazu führt, dass bei der Erstellung eines Team-Objektes im Controller-Test ein Fehler bei der Validierung geschmissen wird. Nach verschiedenen Strategien habe ich noch keine Lösung für dieses Problem gefunden. Falls jemand an dieser Stelle eine Idee hat, bitte kommentieren.

lheilmann:
After consulting PO, we decided to create a new issue in the upcoming sprint for the remaining work on issue #27. The base functionality for roles in team and corresponding tests are already implemented and found in this pull request. The remaining work (criteria 1 of #27) consists of the validation of the associations `team_owner` and `team_member` in the team model and the corresponding test. A first approach is found in the model, but commented. Testing these validations are more difficult and time-consuming, than we thought.

spec/models/team_spec.rb
```diff
@@ -12,7 +12,34 @@
 
 RSpec.describe Team, type: :model do
   it "is valid when produced by a factory" do
-    team = FactoryBot.build(:team)
+    team = FactoryBot.create :team
```
birneamstiel: Is there a reason you need to use `create` and make the team instance persistent? Using `build` would speed up the tests I think...

spec/models/team_spec.rb
```diff
@@ -12,7 +12,34 @@
...
+
+  it "should have and belong to team members" do
+    # After building a team via FactoryBot a member is created using the factory for users and assigned to the team
+    team = FactoryBot.create :team
+    expect(team.members).to have(1).items
+  end
+
+  it "should be able to have multiple team owners" do
+    team = FactoryBot.create :team, :with_multiple_owners
+    expect(team.team_owners).to have(2).items
```
birneamstiel: I think it's not really clear how many team owners a freshly created team has since the trait `:with_multiple_owners` is somewhat ambiguous...
I'd either give the trait a more descriptive name (e.g. `:with_two_owners`) or change the trait to accept the number of owners as parameter.

spec/models/team_spec.rb
```diff
@@ -12,7 +12,34 @@
...
+    expect(team.owners).to have(2).items
+    expect(team.team_members).to have(2).items
+    expect(team.members).to have(2).items
+  end
+
+  it "should be able to have multiple team members" do
+    team = FactoryBot.create :team, :with_multiple_members
+    expect(team.team_members).to have(5).items
```
birneamstiel: see above

spec/models/team_spec.rb
lheilmann: I do not know why I replaced the function - changed back to `build`.

spec/models/team_spec.rb
lheilmann: Implemented the first approach to use more explicit trait names.

birneamstiel (approved):
LGTM
(except for codeclimate of course :-) )

spec/controllers/teams_controller_spec.rb
```diff
@@ -17,6 +17,11 @@
     FactoryBot.build(:team, name: '').attributes
   }
 
+  before(:each) do
+    user = FactoryBot.create(:user)
```
bakoe: Please use consistently either `.create :user` or `.create(:user)`. I'd prefer the option without brackets personally.

spec/controllers/teams_controller_spec.rb
```diff
@@ -107,6 +124,14 @@
       }.to change(Team, :count).by(-1)
     end
 
+    it "deletes the associated team ownerships and team memberships" do
+      team = FactoryBot.create(:team)
```
bakoe: Please use consistently either `.create :team` or `.create(:team)`. I'd prefer the option without brackets personally.

spec/factories/teams.rb
```diff
@@ -17,5 +17,23 @@
...
+  trait :with_two_owners do
+    after(:build) do |team|
+      team.owners = build_list :user, 2
+      team.members = team.owners
+    end
+  end
+
+  trait :with_five_members do
+    after(:build) do |team|
+      team.members = build_list :user, 5
```
bakoe: I think that the trait `:with_five_members` constructs an invalid Team model: It has one owner that **is not a member**. The reason for that is that the team's `members` (that contain the team owner after line 23) are **replaced** by the newly built list of team members in line 36. According to the (implicit?) acceptance criteria that every team owner is also a team member, the resulting Team model is invalid to my mind. 

One solution would be to append the newly built list of team members to the existing team members in line 36:

```ruby
team.members = team.members + build_list :user, 5
```

This would however require to change the "should be able to have multiple team members" test to something like:

```ruby
expect(team.team_members).to have_at_least(5).items
```

spec/models/team_spec.rb
```diff
@@ -12,7 +12,34 @@
...
+    expect(team.team_owners).to have(2).items
+    expect(team.owners).to have(2).items
+    expect(team.team_members).to have(2).items
+    expect(team.members).to have(2).items
+  end
+
+  it "should be able to have multiple team members" do
+    team = FactoryBot.build :team, :with_five_members
+    expect(team.team_members).to have(5).items
+    expect(team.members).to have(5).items
```
bakoe: _This line and the line above: see comment to `spec/factories/teams.rb`_

spec/models/team_member_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe TeamMember, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
bakoe: I think you can remove this empty spec file according to the automatically generated notice (“add some examples to (**or delete**) …”).

spec/models/team_owner_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe TeamOwner, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
bakoe: I think you can remove this empty spec file according to the automatically generated notice (“add some examples to (**or delete**) …”).

bakoe (changes requested):
One possible future problem caused by FactoryBot creating an invalid Team instance; some minor formatting and file hygiene hints.

lheilmann:
@birneamstiel Do you have an idea what to refactor in the remaining codeclimate issue?

bakoe:
In general I have the impression that the “duplication” of every team owner into the team members table causes quite some overhead in the specs, factories and controllers. Maybe we could think about a better way to handle the fact that _team owners are always also team members_ in a future issue.

Apart from that: LGTM! :+1:

spec/controllers/teams_controller_spec.rb
lheilmann: 👍

spec/controllers/teams_controller_spec.rb
lheilmann: 👍

spec/factories/teams.rb
lheilmann: Absolutely correct, I did not see this. Thanks for the hint! 👍

spec/models/team_member_spec.rb
lheilmann: I was thinking that this file you be useful later on, but for now it is definitely cleaner to delete it.

spec/models/team_owner_spec.rb
lheilmann: Same here

spec/models/team_spec.rb
lheilmann: 👍

bakoe:
@lheilmann Regarding the remaining codeclimate issue: You could extract a method out of 

```ruby
# Assign team ownership and team membership to current signed in user who created the team
@team.owners << current_user
@team.members << current_user
```

into something like:

```ruby
assign_user_as_owner_and_member
```

and define a private controller method like:

```ruby
def assign_user_as_owner_and_member (user = current_user)
    @team.owners << user
    @team.members << user
end
```

bakoe (approved):

lheilmann:
@bakoe Good point! For now this seems to be a valid solution. In the middle term we should reconsider the extraction, because #25 is implementing methods, that have kind of the same purpose.

Issue #35, opened 2017-11-12
Create a playing schedule for leagues


MarcelSchmidberger:
**AS A** organizer of an event
**I WANT TO** have a playing schedule automatically generated based on the event configurations
**SO THAT** i know who plays with whom when

**INFO**
Generally a sport event is structured in "Spieltagen" for example the bundesliga. On each "Spieltag" everyone participating in the league has one match against an opponent. A "Spieltag" is not bound to the length of a day and can be arbitrary long. 

**ACCEPTANCE CRITERIA**
- [x] the organizer of an event should be able to access the playing schedule page
- [x] the playing schedule "Spielplan" page should display a large table which shows the "Spielpaarungen" so who plays with whom
- [x] the table should show
  - [x] an colomn which indicates the "Spieltag"
  - [x] both player/team names of the participants (**placeholde with numbers for now**)
  - [x] results ->outcome of the match for example 3:2 in football
  - [x] points -> depending on the results points for each team/player are calculated (no calculation necessary yet, manual input for the beginning)
- [x] for starters all "Spieltage" should be displayed in one big table (will be split up later on)
- [x] the "Spielpaarungen" should be based on "jeder gegen jeden" for starters (later one more match making systems)
- [x] it should be assumed that the max count of permitted players is reached all the time and if not entered is some random number for testing (this will be changed as soon as it is possible to sign up for an event)
- [x] this should work for leagues, tournaments are not necessary yet

MarcelSchmidberger:
After a match each participating party receives points for their result. So if we have team A and team B playing a match with the result 5:10, A receives 0 Points, B 3 Points. If they draw both of them might receive 1 point. Therefore it should be possible to enter points for both parties. (The calculation of the points based on the game result is not part of the task)

MarcelSchmidberger:
The structure of the game-days (Spieltage) needs to be refactored. On each game-day one team/player only has one match. A game day does not define a clear date for a match but rather a time span in which the match should be held. The Matches in the playing schedule should be ordered by game-day.
 

This example should help to clarify matters.

![image](https://user-images.githubusercontent.com/29160369/33269406-db1e3a66-d380-11e7-9d55-1a838858b68f.png)

https://www.ran.de/fussball/bundesliga-spielplan

Askir:
[...]

PR #153, opened 2017-11-27
#35/bruteforce/generate schedule

Askir:
adds a schedule page as discussed in #35


app/models/event.rb
```diff
@@ -45,4 +49,40 @@ def end_after_start
       errors.add(:enddate, "must be after startdate.")
     end
   end
+
+  # Everything below this is leagues only code and will be moved to Leagues.rb once there is an actual option to create Leagues AND Tourneys, etc.
+
+  def add_test_teams
+    self.max_teams.times do |index|
+      self.teams << Team.new(name: "Team #{index}", private: false)
```
erksch: `self.` is not actually neccessary here and you do not use it anywhere else, so please maintain constistent code style.

app/models/event.rb
```diff
@@ -35,7 +39,7 @@ def self.types
   scope :active, -> { where('deadline >= ?', Date.current) }
 
   def duration
-    return if enddate.blank? || startdate.blank?
+    return 10 if enddate.blank? || startdate.blank?
```
erksch: Please remove the 10 here.

erksch (approved):

manner:
index method in MatchesController is missing. Please fix that.

Askir:
Intended behaviour there exists a schedule page for each event now

PR #154, opened 2017-11-27
#19/bruteforce/events view

usame-algan:
adds a bootstrapped version of the new event page


app/models/league.rb
```diff
@@ -1,5 +1,4 @@
 class League < Event
   enum game_modes: [:round_robin, :two_halfs, :swiss, :danish]
-
-  validates :name, :discipline, :game_mode, :max_teams, presence: true
+  # validates :name, :discipline, :game_mode, presence: true
```
Askir: why is this commented and max_teams removed? Pretty sure they should be required

app/models/tournament.rb
```diff
@@ -1,5 +1,5 @@
 class Tournament < Event
   enum game_modes: [:ko, :ko_group, :double_elimination]
 
-  validates :name, :discipline, :game_mode, presence: true
+  # validates :name, :discipline, :game_mode, presence: true
```
Askir: same question as in league.rb

app/controllers/events_controller.rb
```diff
@@ -81,6 +81,12 @@ def set_user
 
     # Only allow a trusted parameter "white list" through.
     def event_params
+      if params.has_key? :league
+        params[:event] = params.delete :league
+      elsif params.has_key? :tournament
+        params[:event] = params.delete :tournament
+      end
```
Askir, line 98: All these lines are only nessecarry aslong as we don't actually create League or Tournament Objects. I think we should change that either before merging this or in one of the next prs, so that other teams also know how to properly use the model.

app/views/events/_form.html.erb
```diff
@@ -1,67 +1,98 @@
...
+      <div class="col-lg-10">
+        <%= f.text_field :duration, id: :event_duration, :disabled => true, :class => 'form-control' %></div>
+    </div>
+    <div class="form-group">
+      <div class="col-lg-offset-2 col-lg-10">
+        <%= f.submit t('events.create_event'), :class => 'btn btn-primary' %>
+        <%= link_to t('.cancel', :default => t("helpers.links.cancel")),
+                    events_path, :class => 'btn btn-default' %>
+      </div>
+    </div>
```
Askir, line 96: this all looks fine to me but @erksch has way more knowledge about front end development than me so he should probably have a look at this.

config/locales/de.yml
```diff
@@ -73,6 +73,18 @@ de:
...
+      description: Beschreibung
+      event_type: Eventtyp
+      game_mode: Spielsystem
+      player_type: Spielertyp
+      max_teams: Anzahl Teams
+      discipline: Sportart
+      deadline: Deadline
+      startdate: Start
+      enddate: Ende
+      duration: Dauer
```
Askir: If it doesn't take too much effort, coould you add these for english aswell? :)

Askir (changes requested):

app/models/league.rb
usame-algan: This caused an issue earlier with double validation errors when submitting the create event form which resolved somehow. Is there any reason to include the same validation here if we already have it in the Event class?

Askir (approved):

PR #157, opened 2017-11-27
#28: Introduce role-based abilities for CRUD actions on teams

### LINKED ISSUE #28: Visibility in team

bakoe:
Closes #28.


spec/factories/abilities.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :ability do
```
lheilmann: Consider deleting unused code. Do you think we will need this factory in future issues?

lheilmann (changes requested):

lheilmann (approved):

PR #164, opened 2017-11-28
46/flexibus/feature user and admin permissions

### LINKED ISSUE #46: Admin
AS an admin
I WANT TO have full access to content of the sport portal
IN ORDER to have possibility to edit and delete  any information about players, teams and events

- Another users (not admins) can only delete and edit own account and events/teams that they created.
- admin can also act as normal user. For example he can create events or teams and also participate in sport events.
- enough to programmatically add **one** admin for now

emanuel-metzenthin:
Closes #46.


lheilmann:
@emanuel-metzenthin In general, I suggest to remove the `creator` association at least in the team scope. Our group created multiple associations between the `User` and the `Team`-Model, called `team_owners`/`owners` and `team_members`/`members`. Semantically the `creator` is one `owner`. So having a `creator` in the team scope would be duplicate functionalty/role. I will point out the specific files/locations, where to adjust the code, in my opinion - tomorrow in the morning. ;-) Please talk to me or one of my group members, if some point is not clear to you yet. Cheers

jkbe:
Migrations don't run through successfully for me, 'table "events" already exists'.
Edit: Running `rails dbconsole` followed by `DROP TABLE events` lead to a successful execution of migrations but this is not a pretty solution...

app/models/ability.rb
```diff
@@ -26,16 +26,20 @@ class Ability
...
+    if user.present?
+      user_id = user.id
+      can :manage, User, id: user_id
+      can :join, Event
+      can :manage, Event, creator_id: user_id
+      can :manage, Team, creator_id: user_id
+      can :create, :all
+      cannot :create, User
+
+      if user.admin?
```
jkbe: Any reason for this nesting?

app/views/events/index.html.erb
```diff
@@ -53,6 +57,8 @@
   </tbody>
 </table>
 
-<%= link_to t('.new', :default => t("helpers.links.new")),
-            new_event_path,
-            :class => 'btn btn-primary' %>
+<% if can? :create, Event %>
```
jkbe: I think we could leave this out because it's not an explicit requirement of the user story.

app/views/teams/index.html.erb
```diff
@@ -20,19 +20,24 @@
         <td>
           <%= link_to t('.show', :default => t("helpers.links.show")),
                       team_path(team), :class => 'btn btn-default btn-xs' %>
+          <% if can? :edit, team %>
           <%= link_to t('.edit', :default => t("helpers.links.edit")),
```
jkbe: Please add tabs.

app/views/teams/index.html.erb
```diff
@@ -20,19 +20,24 @@
...
                       edit_team_path(team), :class => 'btn btn-default btn-xs' %>
+          <% end %>
+          <% if can? :destroy, team %>
           <%= link_to t('.destroy', :default => t("helpers.links.destroy")),
```
jkbe: Please add tabs.

app/views/teams/index.html.erb
```diff
@@ -20,19 +20,24 @@
...
                       :class => 'btn btn-xs btn-danger' %>
+          <% end %>
         </td>
       </tr>
     <% end %>
   </tbody>
 </table>
-
+<% if can? :create, Team %>
 <%= link_to t('.new', :default => t("helpers.links.new")),
```
jkbe: Please add tabs.

config/locales/en.yml
```diff
@@ -51,6 +51,13 @@ en:
         one: "event"
         other: "events"
 
+  unauthorized:
+      manage:
+        all: "Not authorized to %{action} %{subject}."
```
jkbe, line 56: Are these used in the UI? Then we should add strings for German as well.

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    event = Event.new
+
+    ability = Ability.new(@user)
+    ability.should_not be_able_to(:manage, event)
+  end
+
+  it 'should allow admin to crud teams they created' do
+    team = Team.new(creator: @user)
+
+    ability = Ability.new(@user)
```
jkbe: Shouldn't this be `@admin` instead of `@user`?

jkbe: Plus 'should allow admin to crud teams they *didn't* create' or something completely else?

spec/models/user_spec.rb
```diff
@@ -127,4 +127,15 @@
     @relation = User.reflect_on_association(:events)
     expect(@relation.macro).to eq :has_and_belongs_to_many
   end
+
+  it "has the admin attribute" do
+    user = FactoryBot.build(:user)
+    expect(user).to have_attributes(:admin => false)
+  end
+
+  it "has the admin attribute set to true, if it is an admin" do
```
jkbe, line 136: Maybe test for 'has the admin attribute set to false, if it is not an admin' as well?

spec/views/events/index.html.erb_spec.rb
```diff
@@ -30,4 +33,48 @@
     expect(rendered).to have_css('table.table-striped')
   end
 
+  it "doesn't render the new button when not signed in" do
+    render
+    expect(rendered).to_not have_selector(:link_or_button, 'New')
```
jkbe: Please use `t()`, also for following four tests.

jkbe: And also for the tests in event show view spec and teams index view spec.

spec/views/events/show.html.erb_spec.rb
```diff
@@ -34,6 +34,32 @@
...
+
+  it "doesn't render the edit button when the event doesn´t belong to the user" do
+    sign_in @user
+    render
+    expect(rendered).to_not have_selector(:link_or_button, 'Edit')
+  end
+
+  it "doesn't render the delete button when the event doesn´t belong to the user" do
+    sign_in @user
+    expect(rendered).to_not have_selector(:link_or_button, 'Delete')
```
jkbe: `render` is not called

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -11,4 +13,26 @@
...
+
+  it "doesn't render the edit button when it's not your team" do
+    sign_in @user
+    render
+    expect(rendered).to_not have_selector(:link_or_button, 'Edit')
+  end
+
+  it "doesn't render the delete button when it's not your team" do
+    sign_in @user
+    expect(rendered).to_not have_selector(:link_or_button, 'Delete')
```
jkbe: `render` is not called

jkbe (dismissed):
I cannot edit player details (acceptance criteria). When I'm signed in as admin and click on 'edit' button inside the show user view for an arbitrary user, I always end up on the edit page of admin user.

emanuel-metzenthin:
@lheilmann I understand the idea of replacing `creator` with `team owner`. But that contradicts our ticket which says only user that created a team should be able to edit it, not users that were assigned an 'owner' title later on. 
Probably something for the PO's to discuss?

emanuel-metzenthin:
@jak-ing The failing migration came from the dev. Someone has overwritten the `events` table, so unfortunately a database reset is necessary to run that migration.

jkbe:
@emanuel-metzenthin @lheilmann going to forward this to Dima and Svitlana but I think for now as of the requirements, we should keep creator and owner associations separated.

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,14 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+
+  rescue_from CanCan::AccessDenied do |exception|
+    if current_user
+      render :file => "#{Rails.root}/public/403.html", :status => 403, :layout => false
```
fpottbaecker, line 6: Maybe you should use actual erb views here to allow for a translation of the views.

app/controllers/events_controller.rb
```diff
@@ -22,6 +23,7 @@ def edit
   # POST /events
   def create
     @event = Event.new(event_params)
+    @event.creator = current_user
 
     if @event.save
       @event.editors << current_user
```
fpottbaecker: Note: Another team already added the editors collection to the event, maybe every editor should be allowed to edit the event, but only creates should be allowed to mange the editors.

app/models/ability.rb
```diff
@@ -26,16 +26,20 @@ class Ability
...
+    can :read, :all
 
-    user ||= User.new # guest user (not logged in)
-    # All users can only update their own user attributes
-    can :update, User, id: user.id
+    if user.present?
+      user_id = user.id
+      can :manage, User, id: user_id
+      can :join, Event
+      can :manage, Event, creator_id: user_id
```
fpottbaecker: Here you might check whether the events editors include the user.

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  after(:context) do
+    @user.destroy
+    @other_user.destroy
+    @admin.destroy
+  end
+
+  it 'should have admin permissions, if the user is admin' do
+    ability = Ability.new(@admin)
+
+    ability.should be_able_to(:manage, :all)
```
fpottbaecker, line 63: We usually use the expect syntax everywhere else, could you adopt here as well?

app/models/ability.rb
MariusDanner: If there is no user present, we would be calling .admin? on nil, which is not possible. If we readd line 37 user ||= User.new # guest user (not logged in) we can remove this nesting, but that doesn't seem to be useful.

app/views/events/index.html.erb
MariusDanner: It is no part of the written user story, but we orally agreed with @zhamanakov in our scrum meeting that the ticket includes this too. Right @zhamanakov ? But we can remove this, if it is not wanted.

app/models/ability.rb
jkbe: Duh, sure, sorry.

emanuel-metzenthin:
Due to increasingly frequent conflicts with permission check implementations of other teams we will merge this branch into dev including only those changes that have been reviewed already.

We will create a new issue for the functionality of admins being able to edit normal users. (-> #189 )

jkbe (approved):

PR #166, opened 2017-11-28
Fix form stylesheet. Ref #94

### LINKED ISSUE #94: when signing up with an invalid password the password field offsets
As can be seen in the picture below, an invalid password on sign up results in an offset of the password field. This should not happen. 

- [x]  the field should remain aligned. The text "is too short" should 

- [x]   The highlighting of the frame of the password field should be just like the blue highlighting of the Name (in the picture below) field, so following the line of the original text field (with 
rounded edges)
- [x] The highlighting of the frame should be the same red color as the red notification above). 

- [x]  The red background for "password" should be removed

- [x] The text "is too short" should start aligned with the left edge of the password field and when apparent should cause the text field beneath it to move down and maintain the gap to the content above

**Who it looks now:**
image

**Who it should look like(approximately):**
image

fpottbaecker:
This should fix #94.

Example screenshot:
screen shot 2017-11-28 at 21 06 32

### DIFF
--- a/app/assets/stylesheets/scaffold.css
+++ b/app/assets/stylesheets/scaffold.css
@@ -46,10 +46,12 @@ div.actions {
   color: green;
 }
 
-.field_with_errors {
-  padding: 2px;
-  background-color: red;
-  display: table;
+.field_with_errors input,textarea,select {
+    border-color:#ee5f5b;
+}
+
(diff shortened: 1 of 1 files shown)


ErikBrendel (approved):
lgtm

PR #167, opened 2017-11-28
Hotfix/#162/event validation

### LINKED ISSUE #159: Event abilities are neither configured nor used
PR #142 introduced event organizers, users that created an event and are able to edit it.

This is done by **not rendering the form if the user cannot edit the event**. Of course, any malicious user might still manually post to the correct url, event thought he does not have permission to do so.

Also a user that is not signed in can still create an event, which causes [this error](https://swt2-errbit-2017.herokuapp.com/apps/5a031be36caee90014ce8531/problems/5a1c6d90c157e400062df9c5). Because the `current_user`, which is `nil` when no user is signed in, is added to the editors collection, which results in the mentioned error.

As a solution, only logged in user should be allow to create events, and only editors of an event should be allowed to edit events. This should be expressed in the `models/ability.rb` from [cancancan](https://github.com/CanCanCommunity/cancancan/wiki/defining-abilities).

Additionally the event controller should have a `load_and_authorize_resource ...` line comparable to the one in the users controller.

### LINKED ISSUE #162: Event model is not correctly validated to mirror database restrictions.
See [this error.](https://swt2-errbit-2017.herokuapp.com/apps/5a031be36caee90014ce8531/problems/5a1c6f2dc157e4000e2df9c5)

The event model does not validate the presence of the `player_type` attribute, so this causes an error when the model is saved.

fpottbaecker:
This will fix #162.

The PR contains some reformatting of edited files, the actual changes are another test in the event model spec and another validation in the event model.

Note that this will not fix #159, so a user that is not logged in can still try to create an event, which causes an error.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -20,7 +20,7 @@
 #
 
 class Event < ApplicationRecord
-  validates :name, :discipline, :game_mode, presence: true
+  validates :name, :discipline, :game_mode, :player_type, presence: true
   validates :deadline, :startdate, :enddate, presence: true
   validate :end_after_start
   enum player_types: [:single, :team]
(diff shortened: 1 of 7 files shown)


fpottbaecker:
Currently part of code that is modifed far more often

PR #172, opened 2017-11-28
Fixed schedule generation for even number of teams ref #35

Askir:
Just properly implementing round robin :)


Askir:
Could you have a look at this @erksch ? I'm pretty sure it works but maybe check it out yourself and try with a few team sizes. The current version only works for uneven teams :sweat_smile:

app/models/event.rb
```diff
@@ -61,30 +61,29 @@ def add_test_teams
   end
 
   def generate_schedule
-    calculate_gamedays
+    calculate_round_robin
   end
 
   def gamedays
```
erksch: Do you still need this function? Seems like your not using it anymore.

erksch:
What about the uneven numbers of teams?

erksch:
@erksch okay nevermind

erksch (approved):

PR #173, opened 2017-11-28
Event model date tests fixed for edge cases at midnight

Askir:


Askir:
In case anyone needs explanation: One method of Date.tomorrow and Date.today is timezone sensitive and the other one is not which results in them not matching up at midnight.

erksch (approved):

sumpfranze (approved):
sorry it took so long, didn't notice. If I take too long next time tag me in slack or something.

on topic: Good changes

sumpfranze:
I think the entire file is indented with 4 spaces instead of 2. I believe as of the scrum master meeting on friday, the standard is 2 spaces indentation and every developer is supposed to fix it as it is noticed.

Issue #57, opened 2017-11-14
Mail verification: confirmation Emails


MerlindlH:
**As an** admin
**I want** my plattform **to** send my users e-mails when they change their account settings
**In order to** add an additional layer of security / prevent account theft

**Acceptance Criteria**:
- [ ] Confirmation mails on: 
 - changing e-mail and deleting the account
 - command execution after clicking the confirmation links 
- [x] notification mails on changing passwords
- should be expendable if other crucial data is added to the platform

**Tasks**:
- [x] find out how to send mails
- [x] templates
- [x] send notification mails
- [ ] send confirmation mails with working confirmation link

PR #174, opened 2017-11-29
57/swteam/confirmation mails

ErikBrendel:
Only a bit of issue #57 is tackled for now, because the rest of it cannot be done in the sprint it was meant for anymore.
With this PR, translated "Your Email has changed" and "Your password was changed" - notification-mails are introduced, which work using the default devise mailer. Additional logic (like an Email-change-verification-mail) would require a significant amount of more work. That's why we would like to merge this small but nice feature into the `dev` now.


app/views/devise/mailer/email_changed.html.erb
```diff
@@ -0,0 +1,4 @@
+<p><%= t('.greeting', name: @resource.first_name.capitalize) %></p>
+
+<% mail = @resource.try(:unconfirmed_email?) ? @resource.unconfirmed_email : @resource.email %>
+<p><%= t('.body', oldmail: @email, newmail: mail) %></p>
```
fpottbaecker, line 4: Maybe you should add some more descriptive text, such as "Wenn du das nicht selber gemacht hast, solltest du dein Passwort ändern." or a report abuse button (even if it might not do anything right now.

fpottbaecker (commented):
Should you not at least send the emails you created views for?

fpottbaecker (approved):

Issue #33, opened 2017-11-11
Calendar settings for events


MarcelSchmidberger:
**AS AN** organiser of an event
**I WANT TO** I want to select a start date, and a finish date when creating an event
**SO THAT** the users know when an event will start and end

## ACCEPTANCE CRITERIA

- [x] you can enter a start date of an event
- [x] you can enter a duration or a finish date
- [x] the selection of the dates should be possible either by entering the numbers, or by selecting them from a calendar view similar to [https://www.bahn.de/p/view/index.shtml](url)

GittiHab:
Following things should be improved:

- [x] If I click on a date in the calendar, the calendar should be hidden and the date is added into the text box.
- [x] It is probably best if we remove the duration field.

GittiHab:
There is in fact a small bug: If I set the date, then the duration is automatically filled in. If I then change the duration but immediately after click on the end date and then somewhere else, the duration and dates are inconsistent.

PR #175, opened 2017-11-29
Issuenumber5/75/choosedatesofevent

ConstantinLange:


pawelsoundso (commented):
This references #33. If you look at the last update from the POs-requirements, they are not fulfilled here entirely. This means in order for this branch to merge, we should think about deleting the "Duration" input field especially. Since it causes a lot of "not-so-well-looking" Problems in the form and adds little functionality. (Atm it contains "NaN" if no dates are yet entered for start and finish).

config/locales/de.yml
```diff
@@ -68,7 +68,7 @@ de:
 
   events:
     create_event: "Veranstaltung erstellen"
-    choose_event_format: "Veranstaltungsart wählen"
+choose_event_format: "Veranstaltungsart wählen"
```
Jeilef: falsch eingerückt

Jeilef (approved):
nice update of the javascript

PR #176, opened 2017-11-29
23/kickerz/team page

jonasnoki:
Yay Team Page is great #23


lheilmann:
What does the current status look like? Are there open technical TODO's? Please list them in a short bullet list, if they exist, or add reviewers to start review of the PR. Cheers

app/controllers/teams_controller.rb
```diff
@@ -1,14 +1,10 @@
...
-  before_action :set_team, only: [:show, :edit, :update, :destroy, :assign_ownership, :delete_membership, :delete_ownership]
-  # before_action :set_user, only: [:assign_ownership, :delete_membership, :delete_ownership]
-  before_action :owners_include_user, only: [:assign_ownership, :delete_membership, :delete_ownership]
+  before_action :set_team, only: [:show, :edit, :update, :destroy, :assign_ownership, :delete_membership, :delete_ownership, :perform_action_on_multiple_members]
   before_action :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
-  load_and_authorize_resource :team
 
   # GET /teams
   def index
+    @teams = Team.all
```
bakoe: Please either use something like `accessible_by` here or re-introduce the deleted `load_and_authorize_resource :team` line. Otherwise, the read abilities (for private teams) won’t work properly.

app/models/team_user.rb
```diff
@@ -14,8 +14,17 @@ class TeamUser < ApplicationRecord
   belongs_to :user
   before_create :init
 
-
   def init
     self.is_owner ||= false # New team members are by default no team owners
   end
+
+  def assign_ownership
+    self.is_owner = true
```
bakoe: `self.save` not necessary at this point?

app/controllers/teams_controller.rb
```diff
@@ -59,57 +55,76 @@ def destroy
...
 
   private
-    def delete_owner_if_existing
-      if owners_include_user
-        team_owners.delete user
-      end
+    def assign_ownership_to_member
+      authorize! :assign_ownership, @team      
+      @member_to_become_owner = TeamUser.find_by(user_id: user, team_id: @team.id)
+      @member_to_become_owner.update_attribute(:is_owner, true)
```
bakoe: Any reason not to use the TeamUser’s `assign_ownership` method?

app/controllers/teams_controller.rb
```diff
@@ -59,57 +55,76 @@ def destroy
...
+    end
+
+    def delete_ownership_from_member
+      authorize! :delete_ownership, @team  
+      @member_to_become_owner = TeamUser.find_by(user_id: user, team_id: @team.id)
```
bakoe: Please find a better name for this variable – seems like copied from the other method

app/controllers/teams_controller.rb
```diff
@@ -59,57 +55,76 @@ def destroy
...
+      @member_to_become_owner.update_attribute(:is_owner, false)
```
bakoe: Any reason not to use the TeamUser’s `delete_ownership` method?

app/views/teams/show.html.erb
```diff
@@ -18,41 +17,80 @@
...
-        </td>
-      </tr>
+        <tr>
+          <td><%= check_box_tag "members[]", member.id %></td>
+          <td><%= member.id %></td>
+          <td><%= member.first_name%></td>
+          <td><%= member.last_name%></td>
+          <td>
+            <%= link_to t('.email', :default => t("helpers.links.email")),
+                        team_path(@team),
```
bakoe: This url has to be adjusted, doesn’t it? I thought it should point to a mailto link?

app/views/teams/show.html.erb
```diff
@@ -18,41 +17,80 @@
...
+                        :method => 'get',
+                        :class => 'btn btn-default btn-xs' %>
+            <% if (@team.owners.include? current_user) %>
+                <% if (@team.owners.include? member) %>
+                    <%= link_to t('.delete_ownership', :default => t("helpers.links.delete_ownership")),
+                                delete_ownership_team_path(@team, :team_member => member.id),
+                                :method => 'post',
+                                :class => 'btn btn-default btn-xs',
+                                :disabled => !(can? :delete_ownership, @team) %>
```
bakoe: Just a matter of preference: `cannot?` vs `!(can?)`

app/views/teams/show.html.erb
```diff
@@ -18,41 +17,80 @@
...
+    <tr>
+      <td>
+        <%= t('.apply_action', :default => t("helpers.teams.apply_action")) %>
+      </td>
+      <td></td>
+      <td></td>
+      <td></td>
+      <td>
+        <%= link_to t('.email', :default => t("helpers.links.email")),
+                    team_path(@team),
```
bakoe: Same as above: right url?

app/views/teams/show.html.erb
```diff
@@ -18,41 +17,80 @@
...
+                    :method => 'get',
+                    :class => 'btn btn-default btn-xs' %>
+
+        <%= submit_tag t('.assign_ownership', :default => t("helpers.links.assign_ownership")), name: "assign_ownership" %>
+        <%= submit_tag t('.delete_ownership', :default => t("helpers.links.delete_ownership")), name: "delete_ownership" %>
+        <%= submit_tag t('.delete_membership', :default => t("helpers.links.delete_membership")), name: "delete_membership" %>
+      </td>
+    </tr>
+    </tbody>
+    <thead>
```
bakoe: Is it common pracice to have the thead at the table’s bottom?

app/views/teams/show.html.erb
```diff
@@ -18,41 +17,80 @@
...
-<% end %>
-<% if can? :destroy, @team %>
-    <%= link_to t('.destroy', :default => t("helpers.links.destroy")),
-                team_path(@team),
-                :method => 'delete',
-                :data => {:confirm => t('.confirm', :default => t("helpers.links.confirm"))},
-                :class => 'btn btn-danger' %>
-<% end %>
\ No newline at end of file
+<%= link_to t('.edit', :default => t("helpers.links.edit")),
```
bakoe: Any reason why you removed the `can?` checks here? It’s important that e.g. non-team owners can’t delete teams (and thus don’t get the UI button displayed)

bakoe (commented):
Some feedback on the code; haven’t tested in browser yet. In general looks pretty good to me! 🙂

app/controllers/teams_controller.rb
niconomaa: ^^

app/views/teams/show.html.erb
jonasnoki: This is not part of this issue

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
...
-        team_owners.delete user
+    def assign_ownership_to_member
+      authorize! :assign_ownership, @team
+      member_to_become_owner = TeamUser.find_by(user_id: user, team_id: @team.id)
+      unless member_to_become_owner.nil?
+        member_to_become_owner.assign_ownership
       end
     end
 
+    def delete_ownership_from_member
```
birneamstiel: For readability, maybe change the way this method retrieves the member whose onwership should be deleted: Instead of using the `user` method, pass it as argument to this method.

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
...
-      team_owners << user
-      redirect_to @team
-    end
+    assign_ownership_to_member
+    redirect_to @team
   end
 
   def delete_ownership
-    delete_owner_if_existing
+    delete_ownership_from_member
```
birneamstiel: Retrieve user whose ownership should be deleted as parameter and pass it as argument here:
`delete_ownership_from_member User.find(params[:team_member])`

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
...
-    @current_ability ||= Ability.new(current_user, params[:team_member])
+  def perform_action_on_multiple_members
+
+    target_members = params[:members]
+    if target_members.include? current_user.id.to_s
+      target_members = (target_members - [current_user.id.to_s]) + [current_user.id.to_s]
+    end
+
+    target_members.each do |member|
+      params[:team_member] = member
```
birneamstiel: Instead of setting the `:team_member` parameter to member just pass the current member as argument to the desired method (e.g. `delete_membership_from_member member`)

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
 
   # Assigns team ownership to a specific team member
   def assign_ownership
-    unless owners_include_user
-      team_owners << user
-      redirect_to @team
-    end
+    assign_ownership_to_member
```
birneamstiel: Refactor `assign_ownership` logic similar to `delete_ownership` (see other comments)

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
...
   def delete_ownership
-    delete_owner_if_existing
+    delete_ownership_from_member
     redirect_to @team
   end
 
   def delete_membership
-    delete_owner_if_existing
-    @team.members.delete(user)
+    delete_membership_from_member
```
birneamstiel: Refactor `delete_membership` logic similar to `delete_ownership` (see other comments)

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,75 @@ def destroy
...
+      @team.members.delete(user)
+    end
+
     def team_owners
       @team_owners ||= @team.owners
     end
 
     def user
-      @user ||= User.find(params[:team_member])
+      @user = User.find(params[:team_member])
```
birneamstiel: If you're following the approach suggested above, this method would be useless and could be removed, I think.

app/controllers/teams_controller.rb
```diff
@@ -59,55 +55,82 @@ def destroy
...
     redirect_to @team
   end
 
-  def current_ability
-    @current_ability ||= Ability.new(current_user, params[:team_member])
+  def perform_action_on_multiple_members
+    target_members = params[:members]
+    current_user_id_string = current_user.id.to_s
+    if target_members.include? current_user_id_string
+      target_members = (target_members - [current_user_id_string]) + [current_user_id_string]
```
lheilmann: This line logically does not make sense to me. Why are you deleting the current user from the members before you add him in the same way? Could you please explain?

app/models/ability.rb
```diff
@@ -88,10 +88,10 @@ def can_assign_ownership(user)
       end
     end
 
-    def can_delete_membership(team_member, user)
-      can :delete_membership, Team, Team do |team|
+    def can_delete_membership(user)
+      can :delete_membership, Team, Team do |team, team_member|
```
lheilmann, line 96: Way better approach than the one before with the second parameter in `initialize` :+1:

app/views/teams/show.html.erb
```diff
@@ -1,8 +1,21 @@
+<script type='text/javascript'>
```
lheilmann: Consider collecting scripts in the `app/javascripts` directory instead for inserting code snippets in the template. The bigger the application gets, the more confusing it will be to search for javascript code. [Here](http://railsapps.github.io/rails-javascript-include-external.html) you can find some details on how to handle javascripts in rails applications. See **Page-Specific Scripts**. Probably also an issue to discuss on the round table :v:

db/migrate/20171205103643_add_primary_key_to_team_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddPrimaryKeyToTeamUser < ActiveRecord::Migration[5.1]
+  def change
+    add_column :team_users, :id, :primary_key
```
lheilmann, line 3: Why do you need this column?

niconomaa: that's an internal requirement of `ActiveRecord` used to perform updates on the model

app/controllers/teams_controller.rb
niconomaa: done so that `current_user`is last in array and thus last to be performed action on

lheilmann: Alright, thanks! :smile_cat:

db/migrate/20171205103643_add_primary_key_to_team_user.rb
lheilmann: Okay... and you probably need to update in terms of `is_owner` property?

app/views/teams/show.html.erb
jonasnoki: Tried moving the script. Still works but in the console I now get this error:
image

jonasnoki: I tried to do it just like the events people did in the events_new show page. When creating an event i get the sam error. ( Still just in the console but idk...)
image

lheilmann: Okay, interesting! But your javascript code still works after migrating to the new location? Tests are all green? Then please commit and push, so that we can reproduce the issue. :heart_eyes:

conradlempert (approved):

PR #181, opened 2017-11-29
Hotfix/#163/correct user authorization

### LINKED ISSUE #163: Users controller causes nil error on resource authorization
See [this error.](https://swt2-errbit-2017.herokuapp.com/apps/5a031be36caee90014ce8531/problems/5a1d947d46d3bb00089d941f)

When a user that is not signed in tries to access the edit page of an existing user, this error occurs.

No way to fix this is known so far.

fpottbaecker:
This will fix #163.

The issue was caused by a duplicate declaration of routes in the `routes.rb`. Because the user routes were not in a devise scope, a before action of the devise registration controller raised an error.

The issue was resolved by removing the duplicate root and a test was put in place covering the erroneous case.


jcremerius (approved):
Fixed the problem with devise mapping for user edits.

Issue #152, opened 2017-11-27
Fix of Dropdown Menu


PoJo93:
AS a logged in User on a mobile device
I WANT TO not be able to hover 
TO control easily the menu

ACCEPTANCE CRITERIA
When I click on the profile name (drop down menu) on a mobile device I will not see the desktop drop-down menu

manner:
Estimated Size: 2
Actual Size: 2

PR #188, opened 2017-11-30
fix responsive navbar hover and vertical dividers. ref. #152

manner:


nikriek (approved):
Are we using SCSS? If yes, there might be some variables to use instead of hardcoding the screen width. Otherwise LGTM 🚀

manner:
Not that I know of. If we use it in the future `@screen-sm` should work...

manner:
connects with #152

PR #193, opened 2017-12-01
Fix all rubocop offenses #192

### LINKED ISSUE #192: Fix all Rubocop warnings
When I execute the linter on my machine using `$ rubocop`, I get a lot of old warnings and error messages. They don't help me with identifying the new warnings that I introduced on my branch before I push to GitHub and get a 'failed' notification from CodeClimate.

jkbe:
Closes #192.
I used the autocorrect option in RubyMine (probably based on `rubocop -a`) to fix all existing offenses in our project (based on the current Rubocop config file). I also turned on checks for the Ruby 1.9 hash syntax which discourages the use of `=>` in hashes where it's not needed (we agreed on this in today's meeting of Scrum Masters). See [documentation for Style/HashSyntax](http://www.rubydoc.info/gems/rubocop/0.51.0/RuboCop/Cop/Style/HashSyntax) (we use `ruby19`).


spec/views/events/edit.html.erb_spec.rb
```diff
@@ -6,23 +6,23 @@
...
-      :description => "MyText",
-      :owner => FactoryBot.build(:user),
-      :game_mode => "MyString",
-      :max_teams => 20,
-      :player_type => Event.player_types[Event.player_types.keys.sample],
-      :discipline => "MyString",
-      :deadline => Date.new(2017,11,16),
-      :startdate => Date.new(2017,12,01),
-      :enddate => Date.new(2017,12,05)
+                              :name => "MyString",
```
fpottbaecker: While we're at it: Can we turn on [`Style/HashSyntax`](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/hash_syntax.rb), which forces the new hash syntax on symbol key only hashes?

fpottbaecker (changes requested):

spec/views/events/edit.html.erb_spec.rb
jkbe: Yes, you're right. I'll do it.

lheilmann (approved):

fpottbaecker:
And additionally, there is [`Style/StringLiterals`](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/string_literals.rb) set to `single_quotes` (by default) which requires a more consistent quoting style (warning increases offenses by about 600. Do we want that too?

spec/views/events/edit.html.erb_spec.rb
jkbe: @Pottiman it looks to me like `ruby19_no_mixed_keys` from http://www.rubydoc.info/gems/rubocop/0.45.0/RuboCop/Cop/Style/HashSyntax would fit us best. Did you happen to see another option which discourages use of symbols as keys (where hash rockets are allowed again)?

jkbe:
A call for consistent quoting style hasn't come up yet so I wouldn't do that right now. However, we can discuss that in the next Scrum Master Meeting.

spec/views/events/edit.html.erb_spec.rb
fpottbaecker: I think the default (`ruby19`) allows rockets in non symbol key hashes, eg
```ruby
a = {
  :a => 1,
  'b' => 2
}
```
is legal.
Do you want to explicitly prohibit this?:
```ruby
a = {
  a: 1,
  'b' => 2
}
```

fpottbaecker:
And finally, you can enable all Rails checks:
```yaml
Rails:
  Enabled: true
```
I do not know whether this is default already.

jkbe:
@Pottiman do you know what the Rails Cop does? When enabled, Rubocop doesn't find any new offenses.

fpottbaecker:
It should enable the [rails cop department](https://github.com/bbatsov/rubocop/tree/master/lib/rubocop/cop/rails), just to be sure.

Maybe it already was on...

.rubocop.yml
```diff
@@ -51,7 +51,8 @@ Layout/FirstParameterIndentation:
 
 # Use Ruby >= 1.9 syntax for hashes. Prefer { a: :b } over { :a => :b }.
 Style/HashSyntax:
-  Enabled: false
+  Enabled: true
+  EnforcedStyle: ruby19
```
fpottbaecker, line 55: This is default, do we need to explicitly state it?

spec/views/events/edit.html.erb_spec.rb
jkbe: As discussed, let's not do this.

.rubocop.yml
jkbe: It's the style in the config file so far so I would just leave it like this.

jkbe:
@Pottiman about the Rails Cop: Let's leave this out for now as there were no new offenses through this and we don't really know what it does.
This PR's main purpose is to fix all offenses that we had in our project with the existing Rubocop config file and to introduce the new HashSyntax rule that we agreed on in the meeting of Scrum Masters.

fpottbaecker (approved):

PR #194, opened 2017-12-01
Hotfix/extract mailer settings

ErikBrendel:
With this PR, the settings and credentials for the ActionMailer **get extracted to environment variables** for the Production env (these varaibles already exist in `heroku(-dev)`). **Development env (like test env) now doesn't deliver mails.** The content and headers of a mail sent will still get **printed to console**, so you (as a manual tester) can still copy the links to click on from there.

Additionally, this PR includes a **schema change**. It seems like someone changed an existing migration and hasn't updated the schema file. With this change, migrations and schema should be in sync again.

After this PR is merged, I will **change the password for the gmx account**, (in gmx and heroku), so that it cannot be compormised anymore by looking into the git history.


nikriek (approved):
Cool, please change the password ;)

fpottbaecker (approved):

ErikBrendel:
Passwort wurde geändert

PR #195, opened 2017-12-01
Refactor team model to contain only one table of both members and owners

### LINKED ISSUE #190: Refactor team user association
Refactor the current associations of  team and user through team_member and team_owner to a single one called team_user with the attribute `is_owner`.

bakoe:
Closes #190


app/models/team_user.rb
```diff
@@ -0,0 +1,12 @@
+# Class for members of teams. They can be owners with advanced rights.
+# By default, new TeamUsers are no owners.
+class TeamUser < ApplicationRecord
+  belongs_to :team
+  belongs_to :user
+  after_initialize :init
```
birneamstiel: we might want to use a different callback here, maybe `before_create`? 
see: https://stackoverflow.com/questions/1550688/how-do-i-create-a-default-value-for-attributes-in-rails-activerecords-model#comment11008051_4576026

db/schema.rb
```diff
@@ -68,16 +68,11 @@
...
     t.integer "user_id", null: false
-    t.index ["user_id"], name: "index_team_members_on_user_id_and_user_id"
-  end
-
-  create_table "team_owners", id: false, force: :cascade do |t|
-    t.integer "team_id", null: false
-    t.integer "user_id", null: false
-    t.index ["user_id"], name: "index_team_owners_on_user_id_and_user_id"
+    t.boolean "is_owner"
+    t.index ["user_id", "team_id"], name: "index_team_users_on_user_id_and_team_id"
```
birneamstiel, line 77: Why didn't we use both id columns (`user_id` and `team_id`) as index for team_members before, was this just inaccurate?

spec/controllers/teams_controller_spec.rb
```diff
@@ -110,16 +110,10 @@
...
         expect {
           post :create, params: { team: valid_attributes }
-        }.to change(TeamOwner, :count).by(1)
-      end
-
-      it "creates a new TeamMember" do
-        expect {
-          post :create, params: { team: valid_attributes }
-        }.to change(TeamMember, :count).by(1)
+        }.to change(TeamUser, :count).by(1)
```
birneamstiel, line 116: Don't you think we should still test that there is a new team owner created?  Maybe by retrieving the last added instance to the teams table via Team.last and checking its owners? see https://stackoverflow.com/questions/36892591/rspec-test-to-check-the-value-of-created-object

spec/controllers/teams_controller_spec.rb
```diff
@@ -200,11 +194,10 @@
     it "deletes the associated team ownerships and team memberships" do
```
birneamstiel: Maybe also rename the description since we aren't actually deleting any team ownership anymore.

spec/controllers/teams_controller_spec.rb
```diff
@@ -200,11 +194,10 @@
...
       team = Team.create! valid_attributes
       team.owners << subject.current_user
```
birneamstiel: Is this line still needed?

spec/controllers/teams_controller_spec.rb
```diff
@@ -233,78 +226,70 @@
...
       team = Team.create! valid_attributes
-      anotherUser = FactoryBot.create :user
+      another_user = FactoryBot.create :user
 
-      post :assign_ownership, params: { id: team.id, team_member: anotherUser.id }
+      post :assign_ownership, params: { id: team.id, team_member: another_user.id }
       expect(response).to be_forbidden
     end
 
     it 'should not add a team owner twice' do
```
birneamstiel: This test is somewhat senseless now, maybe it would be more useful to check if user is still owner after requesting `:assign_ownership` twice (to make sure there is no toggling happening)?

birneamstiel (changes requested):

db/schema.rb
bakoe: I think there was just an inaccuracy/mistake in the old database table structure which was created using `t.index [:user_id, :user_id]` (which makes no sense in my opinion). 

Maybe @lheilmann and/or @jonasnoki know more?

app/models/team_user.rb
bakoe: Thanks for noticing! 🙂

spec/controllers/teams_controller_spec.rb
bakoe: Done in 043f834

birneamstiel (approved):

PR #204, opened 2017-12-05
Issuenumber5/#85/single player join event

stlutz:
- Refactor tests for event index page and show event page. 
- Add leave functionality to leave buttons on event index page and show event page. 
- Change routing for joining and leaving a event to redirect back to the page the request came from.


app/controllers/events_controller.rb
```diff
@@ -53,16 +53,22 @@ def destroy
   end
 
 
-  # PATCH/PUT /events/1/join
+  # PUT /events/1/join
   def join
-    @event.users << current_user
+    @event.add_participant(current_user)
     flash[:success] = "You have successfully joined #{@event.name}!"
```
sumpfranze: hardcoded string. should be internationalized.

app/controllers/events_controller.rb
```diff
@@ -53,16 +53,22 @@ def destroy
...
-    redirect_to @event
+    redirect_back fallback_location: events_url
+  end
+
+  # PUT /events/1/leave
+  def leave
+    @event.remove_participant(current_user)
+    flash[:success] = "You have successfully left #{@event.name}!"
```
sumpfranze: should also be internationalized.

app/models/event.rb
```diff
@@ -88,4 +88,21 @@ def calculate_gamedays
...
+    participants.delete(user)
+  end
+
+  def has_participant?(user)
+    participants.include?(user)
+  end
+
+  # aliasing to ease renaming in database
+  def participants
+    users
```
sumpfranze: what does ease renaming in database mean? What and why should something be renamed in the database?
you should be able to name the association directly with I believe the class_name: option of the HABTM association

spec/features/event/index_event_spec.rb
```diff
@@ -2,68 +2,106 @@
...
 
-  it "should be possible for a user to see, which event he/she joined" do
-     sign_in @user
-     @singleevent = FactoryBot.create :event, player_type: Event.player_types[:single]
-     visit events_path
-     click_link_or_button(:join_event_button)
-     expect(page).to have_content("Participating")
-  end
+      it "should redirect me to itself when clicking the leave button" do
+        expect(current_path).to eq(events_path)
```
sumpfranze: don't you need to click the leave button before checking the path?

spec/features/event/show_event_spec.rb
```diff
@@ -1,44 +1,90 @@
...
+        click_link(:leave_event_button)
+        expect(current_path).to eq(event_path(@event))
+      end
+
+      it "should not have a join button" do
+        expect(page).not_to have_link(:join_event_button)
+      end
+
+      it "should show that I am participating" do
+        expect(page).to have_content("Participating")
```
sumpfranze: hard coded string which should be internationalized in both the test and the actual page.

spec/features/event/index_event_spec.rb
```diff
@@ -2,68 +2,106 @@
...
+      end
 
-     sign_in @user
-     @singleevent = FactoryBot.create :event, player_type: Event.player_types[:single]
-     @singleevent.deadline = Date.yesterday
-     visit events_path
-     expect(page).not_to have_button(:join_event_button)
-  end
+      it "should show that I am participating" do
+        expect(page).to have_content("Participating")
```
sumpfranze: another hard coded string.

sumpfranze (dismissed):
mostly internationalization but a few other things as well.

sumpfranze:
it is still possible to join events whose deadlines have passed. The tests should be updated and the functionality implemented. Probably cancancan needs to check that.
- can be fixed by only permitting the participate actions on the Events where the active scope is applied.
so Event.active

Also as of our last meeting, there should no longer be a difference between a single user joining and a team joining.

ConstantinLange (approved):
Good job!c But we should discuss, whether we use Date.current or Date.today.

PR #207, opened 2017-12-05
Profile property page improvements for acceptance

citron0xa9:
This PR contains improvements required to be accepted by the PO


app/views/users/edit_profile.html.erb
```diff
@@ -30,7 +30,12 @@
   <div class="form-group row">
     <%= f.label :telegram_username, class: 'col-lg-2 col-form-label' %>
     <div class="col-lg-10">
+      <div class="input-group">
```
nikriek: Intendation looks wrong

nikriek (approved):

app/views/users/edit_profile.html.erb
citron0xa9: done

Issue #189, opened 2017-11-30
Admin can edit users


emanuel-metzenthin:
**As** an admin
**I want** to be able to edit and delete users
**In order** to maintain the website.

**Estimation: M (Actual Size: M)**

This has been part of #46 originally.

PR #208, opened 2017-12-05
Admin can edit users

emanuel-metzenthin:
Issue #189

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -99,8 +99,7 @@ class UsersController < Devise::RegistrationsController
     end
 
     def admin_update_params
-      user_params = params[:user]
-      
+      user_params = params[:user] 
       if user_params[:password].blank?
         user_params.delete(:password)
         user_params.delete(:password_confirmation)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -99,7 +99,7 @@ class UsersController < Devise::RegistrationsController
     end
 
     def admin_update_params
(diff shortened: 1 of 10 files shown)


emanuel-metzenthin:
The code climate error doesn't make sense in my opinion because the password has to be deleted from the params and any extraction into variables or methods would only make the code less readable. (@jak-ing)

jonashering (changes requested):
Regarding the authorization management for admin, this looks very good to me!

Just one small request: Clicking on My Account in the Navbar currently results in an error (links to something like `http://localhost:3000/my/users/edit.1`). Maybe changing the link in `_navbar.html.erb` from `edit_user_registration_path`to `edit_user_path` as you did in some of the other files could solve this. Thanks!

app/controllers/users_controller.rb
```diff
@@ -17,6 +17,35 @@ def show
...
+  def edit
+    @user = User.find(params[:id])
+    authorize! :edit, @user
+  end
+
+  def update
+    @user = User.find(params[:id])
+    authorize! :update, @user
+
+    unless user.admin?
```
chrisma: @user?

MariusDanner: What do you mean with that?

chrisma: I was wondering whether line 29 should read `unless @user.admin?`

MariusDanner: Ah ok, I see.
It is supposed to be current_user I think

jonashering (approved):

PR #210, opened 2017-12-07
141/bruteforce/double round robin support

usame-algan:
Added double round robin support


app/models/event.rb
```diff
@@ -74,7 +74,7 @@ def generate_schedule
   end
 
   def calculate_round_robin
-    pairings_per_day = round_robin_pairings teams.to_a
+    pairings_per_day = round_robin_pairings( teams.to_a)
```
Askir: please don't use braces when they aren't nessecarry

app/models/event.rb
```diff
@@ -85,6 +85,23 @@ def calculate_round_robin
...
 
+  def calculate_double_round_robin
+    pairings_per_day = round_robin_pairings(teams.to_a)
+    pairings_per_day += round_robin_pairings(teams.to_a)
+    pairings_per_day.each_with_index do |day, gameday|
+      day.each do |pairing|
+        # Creating a match for every pairing if one of the teams is nil (which happens if there is an odd number of teams)
+        # the other team will have to wait for this day
+        if gameday < (teams.to_a.size / 2)
+          matches << Match.new(team_home: pairing[0], team_away: pairing[1], gameday: gameday + 1) unless pairing[0].nil? or pairing[1].nil?
```
Askir: the nil? check should be included in the `round_robins_pairings` method. So that it only return valid pairings. That should reduce code duplication

app/models/event.rb
```diff
@@ -85,6 +85,23 @@ def calculate_round_robin
...
   end
 
+  def calculate_double_round_robin
+    pairings_per_day = round_robin_pairings(teams.to_a)
+    pairings_per_day += round_robin_pairings(teams.to_a)
+    pairings_per_day.each_with_index do |day, gameday|
+      day.each do |pairing|
+        # Creating a match for every pairing if one of the teams is nil (which happens if there is an odd number of teams)
+        # the other team will have to wait for this day
+        if gameday < (teams.to_a.size / 2)
```
Askir: you don't need to convert to array to get a size of teams. Also again please remove the braces :)

spec/models/league_spec.rb
```diff
@@ -58,12 +61,20 @@
...
+        expect(gameday_matches.length).to be 2
       end
+    end
+
+    it "double round robin has double the gamedays as normal round robin" do
+      new_league = FactoryBot.create(:league_with_teams)
+      new_league.calculate_double_round_robin
+      new_matches = new_league.matches
+
+      expect(new_matches.length).to be new_league.teams.length * (new_league.teams.length - 1)
```
Askir: you aren't actually checking for gamedays but just for how many matches are created. Maybe change the test to be a bit more precise (or add a few more)

app/models/event.rb
```diff
@@ -85,6 +85,23 @@ def calculate_round_robin
     save
   end
 
+  def calculate_double_round_robin
```
Askir: If I see correctly, this method is never called and so doesn't actually do anything. It should probably be checked what mode is active in the generate_schedule method and create the right pairings.

Askir (changes requested):

app/models/event.rb
```diff
@@ -70,7 +70,11 @@ def add_test_teams
   end
 
   def generate_schedule
-    calculate_round_robin
+    if game_mode == 0
+      calculate_round_robin
+    elsif game_mode == 1
+      calculate_double_round_robin
```
Askir: it would be neat, if you could use the actual enum here for more readability. Also in the new version of the dev this code is in the league.rb instead of the event.rb as you'll see when you merge.

app/models/league.rb
```diff
@@ -29,7 +29,11 @@ def add_test_teams
   end
 
   def generate_schedule
-    calculate_round_robin
+    if game_mode == League.game_modes[:round_robin]
+      calculate_round_robin
+    elsif game_mode == League.game_modes[:two_halfs]
+      calculate_double_round_robin
```
ErikBrendel: Are double round Robin and two halfs the same? Then we should probably use only one of these names in the code

spec/factories/leagues.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :league, parent: :event, class: :league do
-    game_mode League.game_modes[League.game_modes.keys.sample]
+    game_mode League.game_modes[:round_robin]
```
ErikBrendel: I do not really understand what's happening here: you set the `game_mode` of FactoryBot-created leagues to `round_robin`, but you also set it manually to the same value in `league_spec > let(:league) `

usame-algan: The line in `league_spec` was from an earlier version. Should be fixed now.

Askir: Imo there is no need to remove the random selection of a gamemode in the factory. We just have to specify that we want a specific game mode if we test for a specific gamemode.

app/models/league.rb
```diff
@@ -44,6 +48,23 @@ def calculate_round_robin
...
 
+  def calculate_two_halfs
+    pairings_per_day = round_robin_pairings teams.to_a
+    pairings_per_day += round_robin_pairings teams.to_a
+    pairings_per_day.each_with_index do |day, gameday|
+      day.each do |pairing|
+        # Creating a match for every pairing if one of the teams is nil (which happens if there is an odd number of teams)
+        # the other team will have to wait for this day
+        if gameday < teams.size
+          matches << Match.new(team_home: pairing[0], team_away: pairing[1], gameday: gameday + 1) unless pairing[0].nil? || pairing[1].nil?
```
Askir: We can chain a map and select call `round_robin_pairings` method and remove the nil pairings with it. That way we don't have to catch the edge case here anymore.

Askir (approved):

PR #212, opened 2017-12-08
Swteam/#170/tournament overview

Markus-Brand:
Does not fully complete the user-story. There is a non-dynamic table showing the team-standings.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -28,6 +28,8 @@ class Ability
   def initialize(user, team_member = nil)
     can :read, :all
     cannot :read, Team, private: true
+    can :schedule, Event
+    can :overview, Event
 
     if user.present?
       user_id = user.id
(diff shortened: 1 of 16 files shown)


app/models/ability.rb
```diff
@@ -26,6 +26,7 @@ class Ability
   # https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities
 
   def initialize(user, team_member = nil)
+    alias_action :schedule, :overview, to: :read
```
ErikBrendel: what is aliased here? What does this do?

app/views/events/overview.html.erb
```diff
@@ -0,0 +1,21 @@
+<%- model_class = Event -%>
+<div class="page-header">
+  <h1><%= t('events.overview.title') %></h1>
+</div>
+
+<table class="table table-striped">
+  <thead>
+  <tr>
+    <th><%= model_class.human_attribute_name(:team) %></th>
+    <th><%= model_class.human_attribute_name(:platzierung) %></th>
```
ErikBrendel: why not use `t` here?

ErikBrendel (approved):
lgtm

PR #216, opened 2017-12-08
Bruteforce/#128/playing schedule with joined players

CoderWombat:


app/controllers/events_controller.rb
```diff
@@ -69,8 +69,8 @@ def leave
 
   # GET /events/1/schedule
   def schedule
-    if @event.teams.empty?
-      @event.add_test_teams
+    @event = Event.find(params[:id])
```
Askir: before action :set_event already does this, so it isn't necessary.

app/models/event.rb
```diff
@@ -62,18 +62,30 @@ def single_player?
...
   # Everything below this is leagues only code and will be moved to Leagues.rb once there is an actual option to create Leagues AND Tourneys, etc.
-
-  def add_test_teams
-    max_teams.times do |index|
-      teams << FactoryBot.create(:team)
+  # Joining a single Player for a single Player Event
+  # This method is only temporary until we have a working teams-infrastructure
+  def add_single_player_team(user)
+    if teams.length < max_teams
+      teams << Team.new(name: "#{user.first_name} #{user.last_name}"  , private: false)
```
Askir: maybe take the users email as the teamname because that one is assured to be unique, while the name might not be. This is a minor issue though.

app/models/event.rb
```diff
@@ -62,18 +62,30 @@ def single_player?
...
+    matches.delete_all
+  end
+
   def generate_schedule
     calculate_round_robin
   end
 
   def calculate_round_robin
+    puts "hello"
+    puts teams
```
Askir: debug messages right? Please remove

app/models/event.rb
```diff
@@ -82,7 +94,7 @@ def calculate_round_robin
         matches << Match.new(team_home: pairing[0], team_away: pairing[1], gameday: gameday + 1) unless pairing[0].nil? or pairing[1].nil?
       end
     end
-    save
+    self.save
```
Askir: Is there any reason for this? Do our coding standard include using self even when it isn't necessary? Because I don't think we should use it if it isn't needed. And if we do we should at least do it everywhere.

db/schema.rb
```diff
@@ -42,6 +42,8 @@
   create_table "events_users", id: false, force: :cascade do |t|
     t.integer "event_id", null: false
     t.integer "user_id", null: false
+    t.index ["event_id", "user_id"], name: "index_events_users_on_event_id_and_user_id"
+    t.index ["user_id", "event_id"], name: "index_events_users_on_user_id_and_event_id"
```
Askir: Is having an index important here? And if yes, do you mind explaining to me why?^^

Askir (changes requested):
A few minor issues, otherwise looks good to me. I just hope the workarounds don't clash with the work of other teams.

db/schema.rb
laurenzse: This just resulted from merging code from the dev-branch. I also don't know why they've added it. What do you suggest?

app/models/event.rb
laurenzse: I would leave it as-is for now, since this code will definitely be removed once the other teams add the joining-functionality. For our PO, this naming will look better during demoing.

laurenzse:
@Askir Do you have suggestions on how to fix this Code Climate issue?

> EventsController#schedule calls '@event.matches' 2 times

erksch (commented):
Please modify the redirection on the schedule page. If the team is a single user, the users name should be displayed as the team name and clicking on it should redirect to the user page. Otherwise, if the team is just a normal team, the team should be displayed and clicking on the team name should redirect to the team page (already implemented). So you no need to check if the team is a single user team and change the link and caption if true.

spec/controllers/events_controller_spec.rb
```diff
@@ -278,13 +278,18 @@
   describe "GET #schedule" do
     it "should generate schedule if not existing" do
       event = Event.create! valid_attributes
-      get :schedule, params: { id: event.to_param }, session: valid_session
+      event.add_participant(@user)
+      event.add_participant(@other_user)
+      event.generate_schedule
+      get :schedule, params: {id: event.to_param}, session: valid_session
+      puts "Matches:"
+      puts event.matches
```
erksch: Why do you need the `puts`?

Askir:
Once you merge the new version of the master into this, you can also move all the schedule methods to the league.rb file. Make sure to also specify that the schedule tests only work with League models afterwards and we should be gtg.

Askir (approved):

PR #228, opened 2017-12-09
Issuenumber5/#29/Join event as a team

pawelsoundso:
Add the possibility to join and leave events as a team with teams, where the current user is owner of.


sumpfranze:
wasn't this User story supposed to be the one that unifies joining team- and single player events into team-join only?

pawelsoundso:
We decided to develop this version, without the unification, because we depended on some other features like Min/Max Members per team in order to realize the unified version. So in order to fulfill this US before the release of the MVP, we didn't want to wait until these other User Stories are completed and created this (not unified) version of "Join as a team". AFAIK the motivation for this unification is not originating from the feature requested by this US itself, but from our own "team-internal" objectives. This and the fact that feature to the outside remains the same, independent from the implementation (unified or not), let us to believe that this unification could be done as part of later US as well.

Jeilef (dismissed):
The user Story is not done. Our DoD doesnt accept this US. The reason we merge this PR is so that the customer can see that the functionality is already implemented. We expect to invest time in the next sprint to retrieve Absolution for our sins

stlutz:
This will need some more work.

Issue #203, opened 2017-12-05
Add getters and setters for owners and members of Teams and avoid usage of <<, + etc. on the relations


bakoe:
- **Follow-up** for #27 and #190
- **Is related** to #177

Currently, when you add a user in a team's **members** (`team.members << user`) and add the same user as one of the team's **owners** afterwards (`team.owners << user`), the TeamUser association table contains two entries à la:

| user_id  | team_id | is_owner |
| ------------- | ------------- | ------------- |
| 1  | 1  | false  |
| 1  | 1  | true  |

It would be better if operations like `<<`, `+`, `-`, `.delete` and so on would work on `team.owners` in a manner that existing TeamUser entries get updated. This would remove the duplicate entries in the TeamUser association table as a side-effect.

niconomaa:
different from what i said before, using `<<`, `+`, `-`, `.delete` on `team.owners` still messes with `team.members` ... however:

none of methods mentioned above is (and should) be performed on `team.owners`. `TeamsController` only interacts with the `is_owner`-flag in a `TeamUser` entry, as everyone should. 

i dare to doubt that we should alter the methods that only cause harm when used inappropriately (and never from the user interface!)

bakoe:
Clarification: Add getter and setter methods for owners and members on Team model which check if a corresponding TeamUser already exists. If this is the case, the setters do not create new TeamUser entries but update the existing ones using their `assign_ownership` and `delete_ownership` methods.

Additionally, add explanatory comments OR something like NotImplementedError avoiding the usage of `<<`, `+` etc. on Team.owners and Team.members.

Furthermore, make TeamUser's {user_id, team_id} combination a unique key in the database in order to avoid other Developers using `<<`, `+` etc. wrongly.

Finally, replace all occurrences of `<<`, `+` etc. on Team.owners and Team.members  (also in the tests!) by usage of the introduced getters and setters.

PR #229, opened 2017-12-09
Enable team members to invite other users as team members via email

birneamstiel:
Sending an email for confirmation of the invitation is not implemented yet since there is no logic for sending mails implemented in our project.


app/views/teams/show.html.erb
```diff
@@ -43,6 +43,37 @@
...
+
+  <div class="modal fade" id="invite_user_by_email_modal" tabindex="-1" role="dialog" aria-labelledby="<%= t('teams.invite_user_to_team.title') %>" aria-hidden="true">
+    <div class="modal-dialog modal-sm" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-label="<%= t('helpers.links.close') %>"><span aria-hidden="true">&times;</span></button>
+          <h4 class="modal-title"><%= t('teams.invite_user_to_team.title') %></h4>
+        </div>
+        <form action="<%= assign_membership_by_email_team_path(@team) %>" method="post">
+          <%= hidden_field_tag :authenticity_token, form_authenticity_token %>
```
birneamstiel: @bakoe I had to add this line since the CSRF verifcation of rails would fail otherwise. Do you think we could somehow use the form builder by rails for building this form? The form builder automatically adds this hidden field...

For further reading and other possible solutions:
http://guides.rubyonrails.org/security.html#cross-site-request-forgery-csrf
https://stackoverflow.com/questions/3364492/actioncontrollerinvalidauthenticitytoken

app/controllers/teams_controller.rb
```diff
@@ -76,11 +76,31 @@ def delete_membership
...
+    redirect_to @team
+  end
+
   def current_ability
     @current_ability ||= Ability.new(current_user, params[:team_member])
   end
 
   private
+    def assign_membership_to_user (new_member)
+      unless @team.members.include? new_member
```
birneamstiel: @bakoe I wasn't sure on which way of checking if a member is included we agreed on so I chose this simple approach. I could also replace this line by creating a new team_user association and checking if it is included in team.team_members...

app/models/ability.rb
```diff
@@ -61,6 +63,10 @@ def can_crud_team(user_id)
       can :destroy, Team, owners: { id: user_id }
     end
 
+    def can_assign_membership_by_email(user)
```
birneamstiel, line 88: @bakoe I had to rename the ability to match the controller action (otherwise I received a `permission denied` when sending a POST request to `can_assign_membership`). I'm open to discuss which is the best way of naming the action/ability (we could also omit the ...by_email suffix).

app/controllers/teams_controller.rb
```diff
@@ -76,11 +76,31 @@ def delete_membership
...
+    end
+    redirect_to @team
+  end
+
   def current_ability
     @current_ability ||= Ability.new(current_user, params[:team_member])
   end
 
   private
+    def assign_membership_to_user (new_member)
```
birneamstiel, line 94: @bakoe For now I've implemented all checks right here, in teams_controller. Not sure if we want to think about moving those checks somewhere (into the model?)

spec/controllers/teams_controller_spec.rb
```diff
@@ -292,4 +292,36 @@
...
+        post :assign_membership_by_email, params: { id: team.id, email: subject.current_user.email }
+      }.to_not change(team.members, :count)
+    end
+
+    it 'does not succeed when email matches no member' do
+      team = Team.create! valid_attributes
+
+      team.members << subject.current_user
+
+      post :assign_membership_by_email, params: { id: team.id, email: nil }
```
birneamstiel, line 323: @bakoe It's probably not best practice to pass a `nil` email but this seemed more reasonable than making up some random string that won't match any email... But as always: open to discussion :-)

niconomaa: even though i would have to try hard to care less, for consistency reasons: in existing tests for #23  we are facing a similar situation in which the string passed in a test is irrelevant; here we have chosen to just pass a sensible, random string instead of `nil`

app/views/teams/show.html.erb
niconomaa: quite a chunk! might it be advisable to use a partial template (potentially not only in this case) to clear our view?
http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials (_3.4 Using Partials_)

app/controllers/teams_controller.rb
```diff
@@ -76,11 +76,32 @@ def delete_membership
...
+      team_members = @team.members
+      unless team_members.include? new_member
+        team_members << new_member
+        # todo: send mail
```
lheilmann, line 98: How is the status here? Do you implement custom emails or do you share code with @MerlindlH's team? Do they already have implemented custom emails? Do you plan to finish this feature in the current sprint or is this going to be a new issue?

app/controllers/teams_controller.rb
```diff
@@ -76,11 +76,32 @@ def delete_membership
     redirect_to @team
   end
 
+  def assign_membership_by_email
+    user = User.find_by_email params[:email]
+    if user
+      assign_membership_to_user user
+    else
+      flash[:error] = I18n.t('teams.invite_user_to_team.no_user_for_mail')
```
lheilmann, line 84: Nice flash error messages! We should use them more consistently in team scope 👍

app/views/teams/show.html.erb
```diff
@@ -43,6 +43,37 @@
...
+          <div class="modal-body">
+            <div class="form-group">
+              <label for="invited-user-email" class="form-control-label"><%= t('activerecord.attributes.user.email') %></label>
+              <input type="email" class="form-control" id="invited-user-email" name="email">
+            </div>
+          </div>
+          <div class="modal-footer">
+            <button type="button" class="btn btn-secondary" data-dismiss="modal"><%= t('helpers.links.cancel') %></button>
```
lheilmann: Just for consistency reasons: Consider using `btn btn-default`, like we do for other buttons on show page.

spec/views/teams/show.html.erb_spec.rb
```diff
@@ -29,6 +29,32 @@
     rendered.should_not have_content(t("helpers.links.assign_ownership"))
   end
 
+  describe 'invite user to team' do
+    describe 'button and modal' do
+      it 'should be rendered for team members' do
```
lheilmann: Consider implementing separate tests for button and modal here. Different reasons for button and modal can cause rendering issues. While for the button an authorization problem can lead to issues, for the modal a javascript error can be problematic. (Not that important though.)

app/controllers/teams_controller.rb
birneamstiel: As far as I know custom emails hasn't been implemented by @MerlindlH's team yet. Though we could implement it by ourselves this would go beyond the scope of this issue (in my opinion). So I would suggest we consider it as follow up and reestimate on Wednesday.

app/views/teams/show.html.erb
birneamstiel: I totally agree with you in matters of consistency but we will lose some semantic information when using the `btn-default` class. Will pass that decision to @bakoe :-)

app/controllers/teams_controller.rb
lheilmann: The acceptance criterias say: "The invited person receives a notification email". In my opinion we can consider it as follow up and reestimate. I would suggest to talk to PO today and ask for clarification of the acceptance criterias.

app/views/teams/show.html.erb
lheilmann: What do you mean by "semantic information"?

birneamstiel: ...that it's a "secondary button", no "default button" in a semantic sense.

app/controllers/teams_controller.rb
birneamstiel: Yeah you're right regarding the criterias but I'm pretty sure we talked about the pre conditions as "logic for sending custom emails is already implemented" when we estimated this issue... But yes, let's talk with @svitlanakalancha about this.

spec/views/teams/show.html.erb_spec.rb
birneamstiel: Good catch, done!

spec/controllers/teams_controller_spec.rb
birneamstiel: @tomkellog I somehow can't find that test you're talking about, could you add a link?

app/views/teams/show.html.erb
birneamstiel: Yes, good thing to consider... @bakoe what's your opinion?

app/models/ability.rb
bakoe: @birneamstiel Following the YAGNI principle I agree with having only the (quite specific) `(can_)assign_membership_by_email` ability and controller method right now. Whenever a more generic `(can_)assign_membership` ability and/or controller method are needed, one can introduce them by factoring out common behavior. Because at the moment only membership assignment via email is requested, I would leave the naming with `_by_email` suffix.

app/views/teams/show.html.erb
bakoe: @birneamstiel @tomkellog I like both of your ideas regarding the usage of rails' form builder (maybe we can use inspiration from [#29's implementation](https://github.com/hpi-swt2/sport-portal/pull/228/files#diff-c100cc1e113f1117d1d232ed4956f698R1)) and a partial template (see [#29's implementation](https://github.com/hpi-swt2/sport-portal/pull/228/files#diff-c100cc1e113f1117d1d232ed4956f698) as well) for the modal and the form contained in it.

I will look into this today!

app/controllers/teams_controller.rb
bakoe: @lheilmann We talked to @svitlanakalancha about the email notifications **before** implementing the issue – sorry for not including you in the discussion in the first place! I added you to the corresponding Slack channel now. Here is a screenshot of the discussion:

email_notifications

lheilmann: Okay. Thanks for the screenshot! If you could find a solution by talking to @svitlanakalancha, that's perfect. For the next time: Please notify the corresponding reviewers or (better) give updates in the issue. :+1:

app/views/teams/show.html.erb
bakoe: I just noticed that the version of Bootstrap currently used in the sport-portal – version 3 – does not even have the `btn-secondary` class. Sorry for that! 😂

I'll replace the class with `btn-default` (for the cancel button) and leave the `btn-primary` class for the submit button.

app/controllers/teams_controller.rb
bakoe: @birneamstiel I think for the moment this is the right place to put these checks – the model is probably going to get some refactoring regarding the primitive operations on `owners` and `members` anyway (see #203). Maybe this will lead to a model-side check regarding i. e. whether the newly added member is already a member of the team. But as long as this is not handled globally in the model, I think it's a good way to put the checks into the controller as is.

conradlempert (approved):
Great commit! <3<3<33<3<3<3<33<3<3<3

PR #230, opened 2017-12-10
Swteam/#104/tournament form

fpottbaecker:
This visually separates the creation form of tournaments and leagues, while keeping duplication to a minimum.


app/views/events/_form.html.erb
```diff
@@ -1,97 +1,91 @@
-<%= form_for @event, :html => { :class => "form-horizontal event" } do |f| %>
+<%= form_for @event, html:{ class: 'form-horizontal event' } do |f| %>
```
ErikBrendel: is `foo:{ bar` really preferable over `foo: { bar` or `foo: {bar`?

app/views/events/_form.html.erb
```diff
@@ -1,97 +1,91 @@
...
 
-    <%= render partial: "application/validation_errors", locals: {model: @event} %>
+    <%= render partial: 'application/validation_errors', locals: {model: @event} %>
 
     <div class="form-group">
-      <%= f.label Event.human_attribute_name(:name), :class => 'control-label col-lg-2' %>
```
ErikBrendel: why not `human_attribute_name` anymore?

app/views/events/_form.html.erb
```diff
@@ -1,97 +1,91 @@
...
-        <%= f.text_field :max_teams, id: :event_max_teams, :class => 'form-control' %>
+        <%= f.text_field :max_teams, class: 'form-control' %>
         <%= f.error_span(:max_teams) %>
       </div>
     </div>
     <div class="form-group">
-      <%= f.label Event.human_attribute_name(:discipline), :class => 'control-label col-lg-2' %>
+      <%= f.label :discipline, class: 'control-label col-lg-2' %>
       <div class="col-lg-10">
-        <%= f.text_field :discipline, id: :event_discipline, :class => 'form-control' %>
```
ErikBrendel: why are the id's not needed anymore?

app/views/events/new.html.erb
```diff
@@ -1,8 +1,7 @@
 <script src="event.js"></script>
 
-<%- model_class = Event -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => :'helpers.titles.new', :model => model_class.model_name.human.titleize %></h1>
+  <h1><%= t '.title', default: :'helpers.titles.new', model: @event.model_name.human.titleize %></h1>
```
ErikBrendel: is `@event` really a fitting name here? (When i see it correctly, there is a class inside this variable --> maybe call it `@event_type` or `@event_class` instead)

features/step_definitions/content.rb
```diff
@@ -26,6 +26,10 @@
   expect(page).to have_field(name, with: text)
 end
 
+Then /^there should be a dropdown '(.*)' with options '(.*)'$/ do |name, options|
+  expect(page).to have_select(name)#, with_options: options.split(','))
```
ErikBrendel: dead code, and a severe case of it. Either implement the step as one would expect it to work, or re-phrase the step.

spec/features/event/new_event_spec.rb
```diff
@@ -11,42 +11,35 @@
...
-    select "Ko", from: "event_game_mode_tournament"
-    select "Single", from: "event_player_type"
-    fill_in "event_max_teams", with: "5"
-
-    fill_in "event_deadline", with: Date.tomorrow.to_s
-    fill_in "event_startdate", with: (Date.tomorrow + 2).to_s
-    fill_in "event_enddate", with: (Date.tomorrow + 3).to_s
+  it 'should be possible to create a date conditions for an event' do
+    visit new_tournament_path
+    fill_in Event.human_attribute_name(:name), with: 'name'
```
ErikBrendel: is this really good? isn't here `I18n` missing?

Markus-Brand (approved):

spec/views/events/new.html.erb_spec.rb
```diff
@@ -1,36 +1,25 @@
...
-
-      assert_select "input[name=?]", "event[discipline]"
-
-      assert_select "input[name=?]", "event[deadline]"
-
-      assert_select "input[name=?]", "event[startdate]"
-
-      assert_select "input[name=?]", "event[enddate]"
-
+    assert_select 'form[action=?][method=?]', events_path, "post" do
```
ErikBrendel: why the double quoted post?

ErikBrendel (changes requested):

app/views/events/_form.html.erb
fpottbaecker: Because you are supposed to call form helpers with symbols pointing to the classes attributes, they already translate it for you. Also this sets the `for` attribute of the label correctly so capybara can find the field if you give it the name of the field in the current locale. Some browser also provide convenience features based on the `for` attribute.

app/views/events/_form.html.erb
fpottbaecker: Because the form helpers automatically set correct ids. We also used that in the introduction to rails.

app/views/events/new.html.erb
fpottbaecker: The `@event` is the event that is being created in the `new` action of the respective controller. `model_name` is defined on the class and any instance.

spec/features/event/new_event_spec.rb
fpottbaecker: `human_attribute_name` already translates the name, according to its behaviour/docs.

ErikBrendel (approved):

app/controllers/events_controller.rb
```diff
@@ -13,11 +13,12 @@ def index
 
   # GET /events/1
   def show
+    @events = nil
```
sumpfranze: does this do anything? The only place I can find where `@events` is set is in the index action. When the show action is called, a new instance of the controller will be created and `@events` is not initialized anymore.

app/controllers/events_controller.rb
```diff
@@ -85,6 +86,11 @@ def set_event
       @event = Event.find(params[:id])
     end
 
+    # Get the type of event that should be created
+    def event_type
+      params[:type] || Event
```
Askir: I'd remove the `|| Event`. Creating an abstract event will most likely lead to issues further down the road. Otherwise this solution seems pretty damn sweet. I was trying to figure out how to make it clean for multiple hours and ended up just adding case-when statements because I couldn't figure out how to it properly.

app/models/event.rb
```diff
@@ -32,7 +32,7 @@ class Event < ApplicationRecord
   validates :name, :discipline, :game_mode, presence: true
   validates :name, :discipline, :game_mode, :player_type, presence: true
   validates :deadline, :startdate, :enddate, presence: true
-  validates :max_teams, numericality: { greater_than_or_equal_to: 0 } # this validation will be moved to League.rb once leagues are being created and not general event objects
+  validates :max_teams, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
```
Askir: Will sooner or later be moved to the league and tournament models, since a rankinglist doesn't havea maximum amount of teams I think.

app/views/events/_form.html.erb
```diff
@@ -1,61 +1,77 @@
...
-      <%= f.text_field :startdate, id: :event_startdate, class: 'form-control', data: { provide: 'datepicker', 'date-format': "yyyy-mm-dd" } %>
+      <%= f.text_field :startdate, class: 'form-control', data: {provide: 'datepicker', 'date-format' => 'yyyy-mm-dd'} %>
       <%= f.error_span(:startdate) %>
     </div>
   </div>
   <div class="form-group">
     <%= f.label :enddate, class: 'control-label col-lg-2' %>
     <div class="col-lg-10">
-      <%= f.text_field :enddate, id: :event_enddate, class: 'form-control', data: { provide: 'datepicker', 'date-format': "yyyy-mm-dd" } %>
+      <%= f.text_field :enddate, class: 'form-control', data: {provide: 'datepicker', 'date-format' => 'yyyy-mm-dd'} %>
```
Askir: Shouldn't it be possible to also parse a symbol here. like `date-format: ''yyyy-mm-dd"` ? instead of mixing strings and symbols? Or is there a specific reason to use one over the other?

app/views/events/_league.html.erb
```diff
@@ -1,9 +1,9 @@
...
 <div class="form-group">
-  <%= f.label Event.human_attribute_name(:game_mode), :class => 'control-label col-lg-2' %>
+  <%= form.label :game_mode, :class => 'control-label col-lg-2' %>
   <div class="col-lg-10">
-    <%= f.select(:game_mode, League.game_modes.map {|k, v| [k.humanize, v]},
-                 {:include_blank => t('events.new.select_game_mode')},
-                 { :class => 'form-control', :id => 'event_game_mode_league'}) %>
+    <%= form.select(:game_mode, League.game_modes.map {|k, v| [k.humanize, v]},
+                 { include_blank: t('events.new.select_game_mode')},
+                 { class: 'form-control' }) %>
```
Askir: Thanks for fixing these

app/views/events/new.html.erb
```diff
@@ -1,8 +1,5 @@
-<script src="event.js"></script>
-
-<%- model_class = Event -%>
 <div class="page-header">
-  <h1><%=t("events.new_#{@event.class.name.downcase}" ) %></h1>
+  <h1><%= t '.title', default: :'helpers.titles.new', model: @event.model_name.human.titleize %></h1>
```
Askir, line 2: much cleaner!

config/routes.rb
```diff
@@ -1,17 +1,14 @@
 # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 Rails.application.routes.draw do
 
-
-  get '/events/create-from-type', to: 'events#create_from_type', as: 'create_event_from_type'
-
-  resources :events do
+  resources :events, only: [:index, :show, :new, :edit, :update, :destroy] do
```
Askir: You could also use `except:` which might be a little nicer here.

config/routes.rb
```diff
@@ -1,17 +1,14 @@
...
     member do
       put :join
       put :leave
     end
   end
-  resources :leagues, controller: 'events', type: 'league'
-  resources :tournaments, controller: 'events', type: 'tournament'
+  resources :leagues, controller: 'events', only: [:show, :new, :create, :update], type: League
+  resources :tournaments, controller: 'events', only: [:show, :new, :create, :update], type: Tournament
```
Askir, line 11: We could also probably use a subclass specific `:edit` route further down the road but this seems fine for now.

features/step_definitions/navigation.rb
```diff
@@ -0,0 +1,7 @@
+When /^he visits the create event path$/ do
+  visit new_event_path
```
Askir, line 2: How has cucumber been working for you guys?

spec/views/events/new.html.erb_spec.rb
```diff
@@ -1,67 +1,51 @@
 require 'rails_helper'
 
-RSpec.describe "events/new", type: :view do
-  context "new league" do
+RSpec.describe 'events/new', type: :view do
```
Askir, line 3: Now that events/new routes to the create_from_type page we could add a test for that here. And also maybe add that for routing tests.

Askir (commented):
Looks pretty good to me, just some minor comments, Î wont block merging for this^^

I hope merging wasn't too much of a problem, I apologize for disrupting your progress with our merge.

app/views/events/_form.html.erb
```diff
@@ -1,97 +1,91 @@
...
-    <div class="form-group" id="tournament-game-mode">
-      <%= f.label Event.human_attribute_name(:game_mode), :class => 'control-label col-lg-2' %>
-      <div class="col-lg-10">
-        <%= f.select(:game_mode, Tournament.game_modes.map {|k, v| [k.humanize, v]},
-                     {:include_blank => t('events.new.select_game_mode')},
-                     { :class => 'form-control', :id => 'event_game_mode_tournament'}) %>
-      </div>
+      <% case @event.type %>
+      <% when 'League' %>
+          <%= f.label :game_mode, class: 'control-label col-lg-2' %>
```
ErikBrendel: this line seems to be in both `when`s, can it be moved outside?

ErikBrendel (approved):

spec/views/events/new.html.erb_spec.rb
fpottbaecker: This is a view spec, but yes, i should test the controller and routing

PR #237, opened 2017-12-10
Kickerz/#22/delete team

conradlempert:
This branch cant be merged yet, because #23 needs to be merged first. You can review my code tho. (Only 2 last commits are new from this branch)

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -53,7 +53,7 @@ class TeamsController < ApplicationController
       @team.destroy
       redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
     else
-      redirect_to @team, notice: I18n.t('helpers.teams.cant_destroy', resource_name: Team.model_name.human).capitalize
+      redirect_to @team, alert: I18n.t('helpers.teams.cant_destroy', resource_name: Team.model_name.human).capitalize
     end
 
   end
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -46,11 +46,16 @@ class TeamsController < ApplicationController
   # DELETE /teams/1
   def destroy
     # Delete all team ownerships and team memberships associated with the team to destroy
-    team_id = @team.id
-    TeamUser.where(team_id: team_id).delete_all
-    @team.destroy
 
-    redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
+    if not @team.in_event?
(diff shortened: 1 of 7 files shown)


app/controllers/teams_controller.rb
```diff
@@ -46,11 +46,16 @@ def update
...
-    @team.destroy
 
-    redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
+    if not @team.in_event?
+      team_id = @team.id
+      TeamUser.where(team_id: team_id).delete_all
+      @team.destroy
+      redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: Team.model_name.human).capitalize
+    else
+      redirect_to teams_url, notice: I18n.t('helpers.teams.cant_destroy', resource_name: Team.model_name.human).capitalize
```
birneamstiel: Since the desired action fails I would choose the `alert:` instead of `notice:` selector (will change the color of the banner from green to red)

app/models/team.rb
```diff
@@ -36,4 +36,9 @@ def matches
   def has_multiple_owners?
     owners.length > 1
   end
+
+  def in_event?
+    events.length > 0
```
birneamstiel: I think we should use `events.exists?` for readability, see: https://apidock.com/rails/ActiveRecord/Base/exists%3F/class

config/locales/en.helpers.yml
```diff
@@ -43,6 +43,7 @@ en:
       private: "Private"
       apply_action: "Apply action to selected:"
       multiple_confirmation: "You could perform the action on user: "
+      cant_destroy: "%{resource_name} kann nicht gelöscht werden, da es noch an Events teilnimmt."
```
birneamstiel: This is where the english translation goes, I think ;-)

app/models/team.rb
birneamstiel: Also: If you want you could add a simple model test for testing that the `in_event?` check works.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171205103643) do
+ActiveRecord::Schema.define(version: 20171206172721) do
```
birneamstiel: Not so sure if we should merge this file into dev (because you haven't changed anything db relevant), maybe override it when updating your branch onto the latest dev?

birneamstiel (changes requested):
Only some minor things, but in general: LGTM :-)

birneamstiel (approved):

Issue #227, opened 2017-12-09
Two-Step event creation form


GittiHab:
**AS A** creator of an event
**I WANT TO** have a two-step event creation form
**SO THAT** I can select the event type before entering other options and type specific options

## ACCEPTANCE CRITERIA
- [x] When clicking on the "New" Button on the "All Events" page the next screen should ask me which event type I would like to create (League, Tournament or Ranking List)
- [x] After I have selected the event type a page is opened where all the general event options as well as the type specific options are

The event type specific options are implemented in  #197,  #104 and by the team [bruteforce](https://github.com/hpi-swt2/sport-portal/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22team+bruteforce%22+)

GittiHab:
Mockup

![two_step](https://user-images.githubusercontent.com/6453623/33795232-ebab67c6-dcdd-11e7-95ea-eec9ddd87372.jpg)

MarcelSchmidberger:
how does this differ from what is already there? 
The "Event type" field should be removed and the "Spielsystem" field should show the correct content, depending on the event type selected in the "Create Event Page". Is this correct or am I mistaken?

GittiHab:
Why can't I view tournaments anymore? Is this because of this user story?
<img width="1710" alt="screen shot 2018-01-06 at 12 46 31" src="https://user-images.githubusercontent.com/6453623/34639423-6c8796a4-f2e0-11e7-8bcf-141711b5b5c0.png">
<img width="1710" alt="screen shot 2018-01-06 at 12 46 28" src="https://user-images.githubusercontent.com/6453623/34639424-6ca7c352-f2e0-11e7-9182-d4de31c6702d.png">

GittiHab:
The bug does not seem to be caused by this user story, so I will accept it.

Issue #145, opened 2017-11-26
Enter Game-Day length in League configuration


MarcelSchmidberger:
As league organizer
I want to select the length of the game days when configuring the league
In order to adjust the frequency of the matches

This is an addition to #19 

**Acceptance Criteria**

- [ ]  game day length can be in the league creation process
- [ ] is used correctly to calculate the new playing schedule

**Tasks**
- [ ]  Refactoring the Event model and its usage in a way that reduces dependencies between different event types
- [ ]  Add the options to the view page
- [ ]  Save it in the database
- [ ]  Modify the schedule view so that it displays gameday length appropriatly

Askir:
@MarcelSchmidberger How exactly does this have an impact on the playing schedule? Should we display a timeframe for each gameday? Because the schedule generation will have to be the same regardless of how long a gameday actually lasts.

MarcelSchmidberger:
The game-day length has no actual impact on the playing schedule calculation. But the time frame of each game-day should be displayed somewhere, so players know when they have to be done with their matches. 
This could look something like that: (please feel free to change the location of the time frame)

![image](https://user-images.githubusercontent.com/29160369/33422912-8b7ab906-d5b7-11e7-8157-839dbfea0aef.png)

Askir:
I've been doing a larger refactor in the scope of this user story, which ended up being larger than expected because of it. 
While doing the refactor I decided to split up the create event pages into different pages for tournaments and leagues. I asked @MarcelSchmidberger before doing so, but now I'm getting different feedback from different teams. Can I have a clarification on this?

MarcelSchmidberger:
Negative values should not be allowed and there needs to be an upper limit the process crashes when entering 10000000000 for the length

MarcelSchmidberger:
it should also be shown in the event page

Askir:
actual size (including refactoring): L

PR #246, opened 2017-12-12
Issuenumber5/#131/createrankinglist

Jeilef:


app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -67,22 +67,82 @@ $( document ).on('turbolinks:load', function() {
...
+        $(field).find("select").attr("disabled", "disabled");
+        $(field).hide();
+    }
+
+    function enableField(field){
+        $(field).find("select").removeAttr("disabled");
+        $(field).show();
+    }
+
+    function enableDatesAndPlayerTypes(){
```
stlutz: I don't know any JavaScript, but this seems a bit ugly to me. As an idea: Might it be possible to define these fields as a collection and then map the enableField (or disableField) method onto each element?

app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -67,22 +67,82 @@ $( document ).on('turbolinks:load', function() {
...
+        disableField(metric);
+        disableField(initialvalue);
+    }
+
+    function enableMetricAndInitialValue() {
+        var metric = $("#event_metric_div");
+        var initialvalue = $("#event_initial_value_div");
+        enableField(metric);
+        enableField(initialvalue);
+    }
```
Askir: Ooops, this is quite unfortunate, in the scope of #145 we decided to to quite the big refactor of the create progress, and now each event type has their own partial. So all the javascript isn't needed anymore. I talked about this with my PO @MarcelSchmidberger beforehand, so there must have been some misscomunication?

We'll have to figure out what the final call on this is.

app/helpers/events_helper.rb
```diff
@@ -1,2 +1,21 @@
...
+  def league_gamemodes_localize
+    Hash[League.game_modes.map { |key, value| [key, I18n.t("events.gamemode.#{key}")] }]
+  end
+
+  def tournament_gamemodes_localize
+    Hash[Tournament.game_modes.map { |key, value| [key, I18n.t("events.gamemode.#{key}")] }]
+  end
+
+  def event_player_types_localize
+    Hash[Event.player_types.map { |key, value| [key, I18n.t("activerecord.attributes.event.player_types.#{key}")] }]
```
Askir, line 11: Good job on these though for sure, missing localization is always a topic when talking to POs :D

app/models/event.rb
```diff
@@ -29,16 +31,24 @@ class Event < ApplicationRecord
 
   scope :active, -> { where('deadline >= ?', Date.current) }
 
-  validates :name, :discipline, :game_mode, presence: true
-  validates :name, :discipline, :game_mode, :player_type, presence: true
-  validates :deadline, :startdate, :enddate, presence: true
+  validates :name, :discipline,  presence: true
+  validates :deadline, :startdate, :game_mode, :player_type, :enddate,
+            presence: true,
+            if: :validate_not_rankinglist?
```
Askir: These validations should just be in the according models if they aren't needed for a rankinglist. Just move them to the Tournament and League instead of having the extra check. Don't forget updating the Factories aswell though :)

app/models/rankinglist.rb
```diff
@@ -0,0 +1,3 @@
+class Rankinglist < Event
+  validates :discipline, :deadline, :startdate, :enddate, :player_type, presence: false
```
Askir: Just a conceptual question: Does Playertype truely not matter here? Because if it doesn't we should maybe talk about removing the field all together for consitency sake. I'm not sure if a League and Tournament have to have this field. Haven't talked to POs about this yet.
Also why is discipline not needed for a ranking list? Isn't a rankinglist for just one discipline (like chess for example) aswell? Otherwise against we need to remove that one from the events validations and put all these validations into the right models.

app/models/event.rb
```diff
@@ -29,16 +31,24 @@ class Event < ApplicationRecord
...
+
   validates :max_teams, numericality: { greater_than_or_equal_to: 0 } # this validation will be moved to League.rb once leagues are being created and not general event objects
   validate :end_after_start
 
   enum player_types: [:single, :team]
+  enum metrics: [:elo, :win_loss, :true_skill]
```
Askir: Put this enum into the Rankinglist model. It's not really need for any other type of event.

app/models/event.rb
```diff
@@ -29,16 +31,24 @@ class Event < ApplicationRecord
...
+
+
+  def validate_not_rankinglist?
+    self.class.name != "Rankinglist"
```
Askir: This method is not needed if the validation is in the Rankinglist model class instead of here.

app/models/tournament.rb
```diff
@@ -21,4 +21,6 @@
 
 class Tournament < Event
   enum game_modes: [:ko, :ko_group, :double_elimination]
+
+  validates :deadline, :startdate, :enddate, presence: true
```
Askir: good job :) but add this to `league.rb` aswell please, if you end up removing it from `event.rb`

app/views/events/_form.html.erb
```diff
@@ -15,77 +15,103 @@
...
+
   <div class="form-group">
-    <%= f.label :event_type, class: 'control-label col-lg-2' %>
+    <%= f.label :event_type, :class => 'control-label col-lg-2' %>
     <div class="col-lg-10">
-      <%= f.collection_select :type, Event.types, :to_s, :to_s,
+      <%= f.collection_select :type, event_types_enum_localize, :first, :second,
                               {selected: @event.type, include_blank: t('events.new.select_type')},
-                              class: 'form-control', id: 'event_type' %>
+                              :class => 'form-control', :id => 'event_type' %>
```
Askir: Again this whole file should look a lot cleaner in the way we refactored the model and the views in the process of #145.
You should just be able to create another partial file: `_rankinglist.html.erb` and add your additional fields there. If you open up the link `/events/new?type=rankinglist` it should automatically render the correct form.

app/models/event.rb
```diff
@@ -29,16 +31,24 @@ class Event < ApplicationRecord
 
   scope :active, -> { where('deadline >= ?', Date.current) }
 
-  validates :name, :discipline, :game_mode, presence: true
-  validates :name, :discipline, :game_mode, :player_type, presence: true
-  validates :deadline, :startdate, :enddate, presence: true
+  validates :name, :discipline,  presence: true
+  validates :deadline, :startdate, :game_mode, :player_type, :enddate,
```
stlutz: This would make more sense in the Tournament and League subclasses (where those fields are actually required fields).

db/migrate/20171208151635_add_metric_to_events.rb
```diff
@@ -0,0 +1,6 @@
+class AddMetricToEvents < ActiveRecord::Migration[5.1]
+  def change
+    add_column :events, :metric, :integer
```
Askir: game_mode is supposed to handle what you are calling metric right now the tournament saves its tournament type(KO, double-KO) in the column. The League saves the league type(round robin, double round robin, etc.) in that column, so I would suggest that the rankinglist saves what system it uses in that column aswell (Elo, trueskill, or whatever).

db/migrate/20171212084905_change_game_mode_and_player_types_null.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeGameModeAndPlayerTypesNull < ActiveRecord::Migration[5.1]
+  def change
+    change_column :events, :player_type, :integer, :null => true
+    change_column :events, :game_mode, :integer, :null => true
```
Askir: If you end up saving it in the same column, there is no need for this in the migration and also we obviously have to figure out if player_type is gonna stay or not.

app/models/event.rb
stlutz: Checking for the class name could probably be avoided by (as mentioned before) moving the corresponding validation to the relevant subclasses.

app/models/rankinglist.rb
stlutz: Why does the Rankinglist not want a discipline or a player_type? Doesn't Event already check for discipline's presence?

spec/factories/rankinglists.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :rankinglist do
```
Askir: make this a child of event by using `parent: :event` and set the class to rankinglist by `class: :rankinglist` See the tournament and league factory on dev if you want an example. Also if you end up removing the start and enddate from the even factory itself (because it isn't needed for rankinglists) please add it into the tournament and league factories.

app/views/events/_form.html.erb
```diff
@@ -15,77 +15,103 @@
       <%= f.text_area :description, id: :event_description, class: 'form-control' %>
     </div>
   </div>
+
   <div class="form-group">
-    <%= f.label :event_type, class: 'control-label col-lg-2' %>
+    <%= f.label :event_type, :class => 'control-label col-lg-2' %>
```
stlutz: We decided to use the colon notation in favor of the arrow notation for mapping key value pairs in hashes.

config/locales/de.yml
```diff
@@ -116,9 +121,13 @@ de:
     create_event: "Veranstaltung erstellen"
     choose_event_format: "Veranstaltungsart wählen"
     new_league: "Neue Liga"
-    new_tournament: "Neues Tournier"
+    new_tournament: "Neues Turnier"
```
stlutz: Thank you.

Askir (changes requested):
Sadly we actually have some conflicting information from our POs it seems. Which happens if mutliple teams work on similar pages. I hope we can clear this up.

However I also left some general info on the usage of the event model.
We changed how Events are being created in one of our latest commits, you should just be able to add the ranking list in the create method of the controller.
For more information on how to use the event model you can also look at the small article in the wiki that I created: https://github.com/hpi-swt2/sport-portal/wiki/Das-Event-Model

If there are any other questions on this, feel free to ask me. I can also help you merge this once we have a definitive answer from  our POs on how the create event page should actually look like.

config/locales/de.yml
```diff
@@ -136,6 +146,18 @@ de:
...
       points: "Punkte"
+    metric:
+      elo: "ELO"
+      win_loss: "Gewinne-Niederlagen"
+      true_skill: "Wahres Können"
+    gamemode:
+      ko: "KO"
+      ko_group: "Gruppen-KO"
+      double_elimination: "Doppel-KO-System"
+      round_robin: "Runder Robin"
```
stlutz: According to Wikipedia:

> Ein Runder Robin, englisch Round-robin, ist eine Bitt- und Beschwerdeschrift, auch Petition genannt, die von ihren Verfassern und Befürwortern rundherum in einem Kreis unterschrieben worden ist

I don't think that's what was meant. Other translation include "Jeder-gegen-Jeden" or "Rundenturnier". Even just keeping the english original "Round Robin" would have been more explanatory.
#19 states "Jeder-gegen-Jeden" to be the wanted translation.

config/locales/de.yml
```diff
@@ -136,6 +146,18 @@ de:
...
+      two_halfs: "Zwei Halbe"
```
stlutz: The correct plural of 'half' is 'halves'. The correct translation was sadly already defined in #19. Therefore it should be "Hin- und Rückrunde".

spec/factories/rankinglists.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
```
stlutz, line 1: Does this really output a valid Rankinglist? This does not look like it would pass Event's validations.

stlutz (changes requested):
The Event's validation part seems off to me. Reintroduction of colon notation in events' _form has to be reverted. Translation needs to be according to #19.

app/models/rankinglist.rb
stlutz: Btw. tests testing whether it is possible to actually create a Rankinglist would be really nice.

app/assets/javascripts/eventscripts/create_event.js
sumpfranze: We got the same from our PO, however, for the refactoring of the create Event process, a whole other User story #227 exists. So I would suggest to do that refactoring in the context of that User Story.

app/models/event.rb
sumpfranze: :+1:

app/models/event.rb
sumpfranze: I guess you could do it both ways. Moving it to the Tournament and League models you would get a bit of code duplication, this way you have to "disable" the check for ranking lists. But probably with how few subclasses we have, moving it to League and Tournament would be a bit cleaner

app/models/rankinglist.rb
sumpfranze: I know that my PO Pius @GittiHab likes to have a selection in the create event form for single player/team events. That doesn't necessarily mean we need to store that in the database, though. If we start working with min players per team and max players per team fields, we could infer if it is a single player event, if min players and max players are 1. But yeah, that's something we need to reach a conclusion on.

Discipline should probably be available for ranking lists, though.

In general though, this whole line should be removed, since it doesn't matter what is in these columns anyway and the presence: true validation in the events model is already disabled for ranking lists (Can also change by moving it to the models).

app/views/events/_form.html.erb
sumpfranze: something that will probably be done in the process of #227

db/migrate/20171208151635_add_metric_to_events.rb
sumpfranze: that's a nice suggestion

spec/factories/rankinglists.rb
sumpfranze: I dont think it is necessary to remove `:deadline`, `:enddate` and `:startdate` from the events factory, since you could also just set those values to nil in the ranking list values when they don't get validated.

app/views/events/_form.html.erb
Askir: The refactoring has already been done and is merged in dev. It's also been done by team swteam in one of their user stories, so #227 is basically obsolete if we just merge it correctly now.

spec/factories/rankinglists.rb
Askir: Factories are supposed to create the minimal needed version of a model though, so if a rankinglist doesn't need these dates we should remove it from the model as necessary validation and therefore remove it from the factory aswell.
We could however add a subfactory :event_with_dates which has these values and then let :league and :tournament factories inherit from this factory instead of the base one. That way we don't have to set the values to nil in the rankinglist factory.

app/models/event.rb
```diff
@@ -29,10 +30,9 @@ class Event < ApplicationRecord
 
   scope :active, -> { where('deadline >= ?', Date.current) }
 
-  validates :name, :discipline, :game_mode, :player_type, presence: true
-  validates :deadline, :startdate, :enddate, presence: true
+  validates :name, :discipline, :game_mode,  presence: true
   validates :max_teams, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
-  validate :end_after_start, :start_after_deadline
+  validate :end_after_start, :start_after_deadline #in tournament & league?
```
stlutz: Is there a reason this was not moved to the tournament and league subclasses together with the other validations?

Askir: I think moving this validation and the one for max_teams to tournaments and leagues is the best we can do.  
Sure code duplication might not be super pretty but it's probably still cleaner than having a type check in the event model.

app/views/events/forms/_League.html.erb
```diff
@@ -1,9 +1,19 @@
+<%= render "events/forms/Dates", f: form%>
```
Askir, line 1: I don't think you have to put an absolute path here, just Dates should be enough.

spec/features/event/new_event_spec.rb
```diff
@@ -48,6 +49,52 @@
...
+
+    it "should not show a field for defining whether its a team or single player sport" do
+      expect(page).not_to have_field('event_player_type')
+    end
+
+    it "should not show a field for the duration" do
+      expect(page).not_to have_field('event_duration')
+    end
+
+    it "should be possible to create a rankinglist" do
```
stlutz, line 85: This test contains assumptions about the validity of events. Using a factory to fill in the fields could solve that.

spec/models/tournament_spec.rb
```diff
@@ -24,8 +24,8 @@
 describe "Tournament model", type: :model do
 
   it "is valid when produced by a factory" do
```
stlutz, line 26: A similar test would be useful for ranking lists.

app/views/events/forms/_Tournament.html.erb
```diff
@@ -1,8 +1,19 @@
...
+
+<div class="form-group" id="event_game_modes">
   <%= form.label :game_mode, class: 'control-label col-lg-2' %>
   <div class="col-lg-10">
-    <%= form.select(:game_mode, Tournament.game_modes.map {|k, v| [k.humanize, v]},
-                 { include_blank: t('events.new.select_game_mode')},
-                 { class: 'form-control' }) %>
+    <%= form.collection_select :game_mode, tournament_gamemodes_localize, :first, :second,
+                            {selected: @event.game_mode, include_blank: t('events.new.select_game_mode')},
+                            { :class => 'form-control', id: :tournament_game_mode}%>
```
Askir: There are a few cases where oldschool hashing syntax is used with :key => "value" instead of key: "value". Maybe just let rubocop fix this if it's possible to use in views?

config/locales/de.yml
```diff
@@ -142,6 +154,17 @@ de:
...
+      ko: "KO"
+      ko_group: "Gruppen-KO"
+      double_elimination: "Doppel-KO-System"
+      round_robin: "Jeder gegen Jeden"
+      two_halfs: "Hin- und Rückrunde"
+      swiss: "Schweizer System"
+      danish: "Dänisches System"
+      elo: "ELO"
+      win_loss: "Gewinne-Niederlagen"
+      true_skill: "Wahres Können"
```
Askir: I would personally not translate true skill at all, since it's a specific way of calculating the ranking data and nobody knows what "Wahres Können" is representing.

db/migrate/20171212084905_change_game_mode_and_player_types_null.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeGameModeAndPlayerTypesNull < ActiveRecord::Migration[5.1]
+  def change
+    change_column :events, :player_type, :integer, :null => true
```
Askir: I'm still a bit surprised that this isn't important for you guys at all. Single players can only join teams with the player_type :single and teams will only be able to join events with the player_type :team. So if you let this column be 0 the join and leave methods will not work for you anymore. If you want to add both you can just add that option to the enum and change the code in join and leave to permit both if the option is selected.

Otherwise noone will be able to join a rankinglist (maybe this is intended?)

spec/views/events/new.html.erb_spec.rb
```diff
@@ -13,15 +13,15 @@
...
-      expect(rendered).to have_field(Event.human_attribute_name :enddate)
+      expect(rendered).to have_field("league_name")
+      expect(rendered).to have_field("event_description")
+      expect(rendered).to have_select("league_game_mode")
+      expect(rendered).to have_select("event_player_type")
+      expect(rendered).to have_field("league_max_teams")
+      expect(rendered).to have_field("league_discipline")
+      expect(rendered).to have_field("event_deadline")
+      expect(rendered).to have_field("event_startdate")
+      expect(rendered).to have_field("event_enddate")
```
Askir, line 24: I don't know what the current styling agreement is, because these two have just been switched in one of the latest PRs I think.

spec/factories/rankinglists.rb
```diff
@@ -0,0 +1,13 @@
...
+  factory :rankinglist, parent: :event, class: :rankinglist do
+    game_mode Rankinglist.game_modes[Rankinglist.game_modes.keys.sample]
+    initial_value 1.3
+    deadline nil
+    startdate nil
+    enddate nil
+    player_type nil
```
Askir: This is definitely not the cleanest solution to a good rankinglist factory. In my opinion. the :event factory should not specify all the dates and player_type (if you end up actually not using it), but instead have a subfactory `:event_wtih_dates`
that specifies all these values. That way you can inherit from the base factory and can cut out the lines 5-8. To make tournaments and leagues work again they just have to inherit from `:event_with_dates.`

Askir (changes requested):
The factory and validation should be changed before merging this. Also please check if you really don't want to use player_type in a rankinglist. Because this will disable the joining of single players AND teams. So I'm not sure if that is intended.

Otherwise: great job on the merge! I definitely thought I'd have to comment on more than this :D

spec/factories/rankinglists.rb
sumpfranze: Well, if the dates are required in the events model, but not in the ranking list model, then setting them in the event factory and setting them to nil in the ranking factory would be the best option in my opinion. That way there are only required values initialized by the factory and the others are nil.

If the validations get moved to the submodels, then I agree, the factories should also display that movement.

stlutz (approved):

Askir:
Codeclimate issue is intended to let people, who implement subclasses of event know what to expect as variable.

Askir (approved):

PR #251, opened 2017-12-12
Event refactor to reduce dependency between different event types

Askir:


erksch (approved):
Looks good to me! A ton of changes but I can see the importance.

PR #254, opened 2017-12-13
Swteam/#169/tournament schedule

ErikBrendel:
Tournaments now have a schedule. This schedule represents the binary tree model behind a ko-tournament.

Codeclimate Issues were already fixed as much as possible. **All open issues should get discussed** with the reviewer. I am not sure how/if these should get fixed.


ErikBrendel:
For some reason, the tests fail on travis. They work locally (and should work in general). Will be further investigated

app/models/match.rb
```diff
@@ -25,4 +25,50 @@ def matches
...
+    I18n.t('matches.round_name.' + key, round: (gameday + 1).to_s, gameid: index.to_s)
+  end
+
+  def has_winner?
+    points_home != nil && points_away != nil && points_home != points_away
+  end
+
+  def winner
+    if has_winner?
+      points_home > points_away ? team_home_recursive : team_away_recursive
```
Markus-Brand, line 53: Does having more points always make you the winner?
In Golf, having a low score is better.
Maybe create a new method for this, so it can be changed more easily later

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171206172721) do
+ActiveRecord::Schema.define(version: 20171212150919) do
```
Markus-Brand: Schema schouldn't get commited

app/models/tournament.rb
```diff
@@ -21,4 +21,120 @@
...
+    set_match_indices
+  end
+
+  def max_match_level
+    Math.log(teams.length, 2).ceil - 1
+  end
+
+  private
+
+    def get_standing_of(team, match)
```
Markus-Brand: I am not happy with the naming of this method, because of the second parameter.
Before having a closer look, i wondered how a team could have a standing in a match.
Maybe rename to lastMatchOf(team, currentMatch) and move the standing_string call to standing_of(team) method?

Markus-Brand (changes requested):

app/models/match.rb
ErikBrendel: yes

app/models/event.rb
```diff
@@ -82,4 +82,10 @@ def can_leave?(user)
   def standing_of(team)
     'Gewinner ' + team.id.to_s
   end
+
+  def add_test_teams
```
DanielStephan97: NEIN!
Das gehört doch in eine Test-Datei? Zumindest nicht in das Event-Model.

app/models/match.rb
```diff
@@ -25,4 +25,50 @@ def matches
   belongs_to :team_home, polymorphic: true
   belongs_to :team_away, polymorphic: true
   belongs_to :event, dependent: :delete
+
+  def name
+    winner_team = winner
+    if winner_team != nil
```
DanielStephan97: Why don't you use present? Or blank?

app/models/match.rb
```diff
@@ -25,4 +25,50 @@ def matches
...
+  end
+
+  def round
+    key = { 0 => 'zero', 1 => 'one', 2 => 'two', 3 => 'three' }[depth]
+    key ||= 'other'
+    I18n.t('matches.round_name.' + key, round: (gameday + 1).to_s, gameid: index.to_s)
+  end
+
+  def has_winner?
+    points_home != nil && points_away != nil && points_home != points_away
```
DanielStephan97: Why don't you use present instead of nil checks?

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
   enum game_modes: [:ko, :ko_group, :double_elimination]
+
+  def standing_of(team)
+    last_match = last_match_of team, finale
+    standing_string team, last_match
+  end
+
+  def finale
+    finale = nil
+    matches.each do |match|
```
DanielStephan97: Use matches.where or better: matches.find

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
+    def standing_recursion_step(team, child)
+      if child.is_a? Match
+        last_match_of team, child
+      end
+    end
+
+    def filled_teams
+      filled_teams = shuffled_teams
+      insert_index = 0
+      until is_power_of_two? filled_teams.length
```
DanielStephan97: This whole loop does not look good. You might just want to split the array of teams in half recursively, consistently ceiling or flooring on each layer and creating matches between those teams.

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
+      filled_teams
+    end
+
+    def shuffled_teams
+      team_copy = []
+      teams.each { |team| team_copy << team }
+      team_copy.shuffle!
+    end
+
+    def build_matches(team_array, depth, index)
```
DanielStephan97: Consistent naming: you are creating matches here.

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
+
+      left_half, right_half = split_teams_array team_array
+      child_depth = depth - 1
+      match_left = build_matches left_half, child_depth, (2 * index)
+      match_right = build_matches right_half, child_depth, (2 * index + 1)
+
+      create_match match_left, match_right, depth, index
+    end
+
+    def split_teams_array(team_array)
```
DanielStephan97: Is there no predefined split method? Or have you tried using `first(n)` and `last(n)`

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
+    def create_match(team_home, team_away, depth, index)
+      match = Match.new team_home: team_home, team_away: team_away, gameday: depth, index: index + 1, event: self
+      match.save!
+      match
+    end
+
+    def beautify_match_indices
+      first_match = matches[0]
+      first_gameday_index_offset = first_match.index - 1
+      matches.each do |match|
```
DanielStephan97: You might want use a hash while creating matches. With something like layer -> array of matches on that layer and add indices based on that.

app/models/tournament.rb
```diff
@@ -21,4 +21,117 @@
...
+    def build_leaf_match(team_home, team_away, depth, index)
+      unless team_home.nil? || team_away.nil?
+        return create_match team_home, team_away, depth, index
+      end
+      team_home || team_away
+    end
+
+    def create_match(team_home, team_away, depth, index)
+      match = Match.new team_home: team_home, team_away: team_away, gameday: depth, index: index + 1, event: self
+      match.save!
```
DanielStephan97, line 105: Maybe you shouldn't raise errors like this. Maybe return a nil on error.

app/views/matches/show.html.erb
```diff
@@ -3,15 +3,29 @@
   <h1><%=t '.title', :default => model_class.model_name.human.titleize %></h1>
 </div>
 
+<%
+def participant_link(participant, participant_type)
+  if participant
+    if participant_type == 'Match'
```
DanielStephan97: Maybe move this method to the events helper.

app/models/event.rb
ErikBrendel: Oh yes, about that method... :D

It's bad, I know. Nobody should use the factory bot in production code. It also produces weird errors when re-starting the server (with events already in the database) and creating a new event.

But: it is really helpful. I do not want to create 16 fake email accounts and register 16 times on the platform to test the creation of event schedules by hand. And it was already there, I am not the original author of this method.

Now that I think of it, we could try to find a better place for calling this method.

app/models/tournament.rb
ErikBrendel: I'm not sure about that: having nil as match in the tournament-match-tree only results in crashes when fetching / displaying the data. schedule and overview - page both rely on not having nils in the tree, only matches and teams. Seeing the error here probably helps finding the problem

app/models/tournament.rb
ErikBrendel: yes, this works better. I changed it

app/models/tournament.rb
ErikBrendel: Do you mean `Match.where` and `Match.find` ?
`matches` is already the array of Match-objects that belong to this event. It seems like this Array-ish object doesn't know the find method.

I tried: `matches.find :first, conditions: 'depth = 0'`

Do you know how to do this?

app/models/tournament.rb
ErikBrendel: Yes, this also sounds interesting. I just think that this would bloat the code unnecessarily: create a (rather) complex data structure for the tree of matches, that already exists as tree (via relation pointers) and as list (in `self.matches`), just for giving these object fitting indices. These indices can be calculated while creating the tree. The only problem is the probably incomplete leaf layer of the tree, that gets adjusted in this method.

app/models/tournament.rb
ErikBrendel: We already thought about that. The problem: This would result in uniformly distributed leaf matches, not all packed to the end of the gameday. The POs wanted them packed in the end, so that every team playes at least once before leaf-match-players play again

Markus-Brand (approved):

PR #258, opened 2017-12-13
Enable rubocop for spec files

jkbe:
Rubocop currently doesn't perform coding style checks on our spec files. However, as we are practicing TDD, tests are an important part of our work and therefore should be checked for our coding style as well (as discussed in meeting of Scrum Masters).


spec/controllers/teams_controller_spec.rb
```diff
@@ -308,17 +308,17 @@
...
+        team.owners << another_user
+        team.owners << still_another_user
 
-      expect {
-        post :perform_action_on_multiple_members, params: { id: team.id, members: [another_user.id, still_another_user.id], delete_ownership: "delete_ownership"}
-      }.to change(team.owners, :count).by(-2)
+        expect {
+          post :perform_action_on_multiple_members, params: { id: team.id, members: [another_user.id, still_another_user.id], delete_ownership: "delete_ownership" }
+        }.to change(team.owners, :count).by(-2)
       end
```
fpottbaecker: this looks to be indented too much

fpottbaecker (commented):
Did you discuss the matter of quotation consistency (suggested: use single quotes unless interpolation or escaping is required [this is rubocop default])?

spec/controllers/teams_controller_spec.rb
jkbe: Yup, just fixed this. Weird autocorrect.

jkbe:
@Pottiman I believe we have not yet discussed this although I'd prefer a uniform style here as well. I'll put it on the agenda for Friday.

fpottbaecker (approved):
Ideally i would want to put off merging until Friday, once the quotation style is chosen

jkbe:
Hm, I think we can also change the configuration step by step, doesn't make that big of a difference.

jkbe:
@hpi-swt2/scrum-masters fyi

PR #260, opened 2017-12-14
Merge to master: Sprint 2

FJThiel:
The POs would like to give the customers something to play around with for the presentation, and the current state of the Master is broken anyway, it can't get worse and will make the POs happy. Everyone wins.


jkbe:
Code Climate always seems to crash / stay in "pending" mode with these big PRs. I'm gonna turn it off as a requirement for merging and will merge, then turn the checks back on for all PRs.

jkbe (approved):

Issue #252, opened 2017-12-12
Team Logo


svitlanakalancha:
As User I want to place **logo** of my team to the **team page** to differentiate from other teams.

**Acceptance Criteria**

- There is a button "Upload logo" on the team page
- Each team member can upload a picture
- Picture can be changed or deleted anytime
- The picture should not be larger than 300 KB
- The picture should be resized to squared form 350 x 350 

**NOTE**
Check existing solution #39

conradlempert:
**Development Tasks**

- [x] Add avatar to team model
- [x] Add upload segment to form view
- [x] Show avatar on team page
- [x] Make picture resolution 350x350
- [x] Allow only team members to change picture with cancancan
- [x] Test Coverage 99.9%

lheilmann:
@svitlanakalancha Please see #252 and talk to @conradlempert about defining the acceptance criteria resizing images.

svitlanakalancha:
we have agreed on it with @conradlempert

PR #263, opened 2017-12-15
#252/kickerz/team logo

conradlempert:


app/views/teams/_form.html.erb
```diff
@@ -27,6 +27,17 @@
     </div>
   </div>
 
+  <div class="form-group row">
+          <%= f.label :avatar, t('activerecord.models.image.one'), :class => 'col-form-label col-lg-2'%>
```
ErikBrendel: When I recall correctly, we agreed to not use the rocket Syntax.

conradlempert: in unserem code ist die rocket syntax halt 100 mal. Ich änder das jetzt mal in meinem code für den pr, aber wenn wir das generell ändern wollen, müssen wir das fett refactoren.

app/views/teams/show.html.erb
```diff
@@ -14,6 +14,7 @@
 
 <div class="page-header">
   <h1><%= t '.title', :default => model_class.model_name.human.titleize %></h1>
+  <%= image_tag @team.avatar_url || image_path('missing_avatar.png'), id: "preview", style: "width:350px; height:350px"%>
```
usame-algan: Would it be better here to use `max-width` or `max-height` to preserve the aspect ratio of the image?

app/models/team.rb
```diff
@@ -9,6 +9,7 @@
 #  description   :text
 #  kind_of_sport :string
 #  private       :boolean
+#  avatar_data   :text
```
usame-algan, line 12: Is there a difference between `avatar_data` and `:avatar`?

conradlempert: Yes

app/views/teams/show.html.erb
conradlempert: The issue says: "The picture should be resized to squared form 350 x 350. " Sounds to me like it should be stretched to that form.

lheilmann: @svitlanakalancha What do you think about resizing the image? In my opinion stretching the image is not the best solution.

app/views/teams/_form.html.erb
```diff
@@ -27,6 +27,17 @@
...
+  <div class="form-group row">
+          <%= f.label :avatar, t('activerecord.models.image.one'), class: 'col-form-label col-lg-2'%>
+          <div class="col-lg-10">
+            <%= image_tag @team.avatar_url || image_path('missing_avatar.png'), id: "preview", style: "max-width:100px; max-height:100px"%>
+            <%= f.file_field :avatar, class: "form-control-file" %>
+            <% if @team.avatar %>
+                  <%= t('helpers.submit.delete', model: t('activerecord.models.image.one'))+ "  " %><%= f.check_box :remove_avatar %>
+              <% end %>
+    </div>
+    </div>
```
lheilmann: Please double-check html structure.

- Label "Bild" is not formatted correctly
- Checkbox is not visible

screen shot 2018-01-15 at 14 49 53

db/migrate/20171215131703_add_avatar_data_to_teams.rb
```diff
@@ -0,0 +1,5 @@
+class AddAvatarDataToTeams < ActiveRecord::Migration[5.1]
+  def change
+    add_column :teams, :avatar_data, :text
+  end
+end
```
lheilmann, line 5: Great!

lheilmann (changes requested):
Some minor issues mainly in views. Please double-check and talk to PO before merging is possible. Rest looks good to me! :+1:

lheilmann (approved):

PR #264, opened 2017-12-16
19/bruteforce/events-view with enum translations

usame-algan:
Added enum translations Ref:#232

PR #265, opened 2017-12-16
Swteam/#244/event show

fpottbaecker:


app/models/event.rb
```diff
@@ -56,7 +56,7 @@ def deadline_has_passed?
   end
 
   def single_player?
-    player_type == Event.player_types[:single]
+    player_type == 'single'
```
ErikBrendel: Is a string 'the way to go' when checking enum equality?

app/models/event.rb
```diff
@@ -82,4 +82,22 @@ def can_leave?(user)
...
+    self.class.human_game_mode game_mode
+  end
+
+  class << self
+    def human_player_type(type)
+      I18n.t("activerecord.attributes.event.player_types.#{type}")
+    end
+
+    def human_game_mode(_)
+      'None'
```
ErikBrendel: Can you somehow annotate that this method should get overridden by subclasses?

app/models/league.rb
```diff
@@ -55,4 +55,8 @@ def round_robin_pairings(teams_array)
       [[teams_array.first, pivot]] + (1...(n / 2)).map { |j| [teams_array[j], teams_array[n - 1 - j]] }
     end
   end
+
+  def self.human_game_mode(mode)
+    I18n.t("activerecord.attributes.league.game_modes.#{mode}")
```
ErikBrendel: Why did you change `game_modes` enum to singular, but use the plural in the translation key?

app/views/events/show.html.erb
```diff
@@ -1,46 +1,68 @@
...
-        <% end %>
-      </span>
-    <% end %>
+    <%=t '.title', default: @event.model_name.human.titleize %>
   </h1>
+  <% if user_signed_in? %>
+    <span class="pull-right">
+      <% if current_user.organizing_events.include?(@event) %>
+        <% if can? :edit, @event %>
+          <%= link_to t('.edit', default: t("helpers.links.edit")),
```
ErikBrendel: Why the double quotes here?

app/views/events/show.html.erb
```diff
@@ -1,46 +1,68 @@
...
+    </span>
+  <% end %>
 </div>
 
 <dl class="dl-horizontal">
-  <% [:name, :description, :discipline, :game_mode, :max_teams, :deadline, :startdate, :enddate].each do |attribute| %>
-      <dt><strong><%= model_class.human_attribute_name(attribute) %>:</strong></dt>
-      <dd><%= @event.send attribute %></dd>
-  <% end %>
+  <dt><strong><%= model_class.human_attribute_name(:name) %>:</strong></dt>
```
ErikBrendel, line 26: That really looks bad. Please do something about that duplication

features/step_definitions/navigation.rb
```diff
@@ -5,3 +5,12 @@
 When /^he visits the new tournament path$/ do
   visit new_tournament_path
 end
+
+
```
ErikBrendel: Space?

features/step_definitions/tournament.rb
```diff
@@ -23,3 +23,8 @@
 When(/^the tournament overview page for (.*) is visited$/) do |tournamentName|
   visit event_overview_path (tournament_named tournamentName)
 end
+
+
```
ErikBrendel: Space?

ErikBrendel (changes requested):
A really nice PR size :)

app/models/event.rb
fpottbaecker: http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

According to docs, the enum attribute returns a string. Symbols actually result in false comparisons

app/models/league.rb
fpottbaecker: Because rails magic. The enums actually had the *wrong name the whole time*. However, rails automatically creates a corresponding plural method that returns all enum constants. Since this is the set of all game modes, i decide to use the plural for the translation keys

app/views/events/show.html.erb
fpottbaecker: I just changed indentation, will fix.

features/step_definitions/navigation.rb
fpottbaecker: the final frontier

app/views/events/show.html.erb
fpottbaecker: there is no nice way to do this. Views inherently have some amount of duplication, just look at the form fields for every model. The previous solution was arguably wrong, since it used `send` and couldn't handle injecting translations. I would ask you to provide a better solution.

features/step_definitions/tournament.rb
fpottbaecker: these are the voyages.

ErikBrendel:
This pr seems to highly overlap with https://github.com/hpi-swt2/sport-portal/pull/264 - you authors should probably discuss this to prevent merge conflicts before they occur

app/models/event.rb
ErikBrendel: When I see correctly, there should be an `event.single?` method already

app/views/events/show.html.erb
ErikBrendel: Just define a function that takes two parameters: the attribute (`:name`) and the value (`@event.name`) and call it 9 times

ErikBrendel (approved):

app/models/event.rb
fpottbaecker: correct

app/views/events/show.html.erb
fpottbaecker: I'm not sure if you can just define a method like that in the respective helper...
Otherwise, it would still look ugly:
```erb
<% vals = {
  name: @event.name,
  discipline: @event.discipline,
  ...
  game_mode: @event.human_game_mode,
  ...
} %>
<% vals.each do |k, v| %>
  <dd><strong><%= model_class.human_attribute_name k %>:</strong></dd>
  <dt><%= v %></dt>
<% end %>
```

And thats not a nice piece of view code. I prefer the style of form views.
I would leave it as it is, since other loops depend on the fact that there will only be this format forever. This is the first time the previous pattern was broken (and i think it's not the last time).

fpottbaecker:
Reviewed and accepted by @MerlindlH 

Pending because of conflicts with #264

fpottbaecker:
This also resolves one criterium of #232

app/views/events/_form.html.erb
```diff
@@ -35,7 +35,7 @@
   <div class="form-group">
     <%= f.label :player_type, class: 'control-label col-lg-2' %>
     <div class="col-lg-10">
-      <%= f.select(:player_type, Event.player_types.map {|k, v| [k.humanize, v]},
+      <%= f.select(:player_type, Event.player_types.map {|k, _| [Event.human_player_type(k), k]},
```
usame-algan: The values of the dropdown options should be 0, 1, 2, ... with the current enums that would be the value and not the key.

usame-algan (approved):

app/views/events/show.html.erb
ErikBrendel: Maybe you are right. Keep it like that if you like

ErikBrendel (approved):

features/events/show.feature
```diff
@@ -0,0 +1,15 @@
...
+    Given a tournament with gamemode ko
+    When the tournament's page is visited
+    Then the page header should show 'Tournament'
+    And the page should show 'K.-o.-System'
+    And the page should show 'Teamteilnahme'
+
+  Scenario: League
+    Given a league with gamemode round_robin
+    When the league's page is visited
+    Then the page header should show 'League'
```
Markus-Brand, line 13: Why is "League" english but "Jeder gegen Jeden" and "Teamteilnahme" german? Please decide for one (English would fit the rest of the sentence)

Markus-Brand (approved):
lgtm, apart from the denglish cucumber tests

PR #266, opened 2017-12-16
Swteam/fix deprecation

fpottbaecker:
This still leaves a few warnings which come from `cucumber-rails` using deprecated tagged hooks.


spec/helpers/events_helper_spec.rb
```diff
@@ -11,5 +11,4 @@
 #   end
 # end
 RSpec.describe EventsHelper, type: :helper do
-  pending "add some examples to (or delete) #{__FILE__}"
```
ErikBrendel: You did none of the two options that this line gave you. Are you sure that having an empty spec is good?

Markus-Brand (commented):
What issue does this refer to?
Why did you to this?

fpottbaecker:
@Markus-Brand No issue.
Because this nuisance came up in PR #259 and it was a fix of 10 minutes.

ErikBrendel (approved):

Issue #234, opened 2017-12-10
Create Sample User Data


PoJo93:
AS a logged in User
I WANT TO see some sample users who are already in the system per default
TO play with the system and try out its functionality

ACCEPTANCE CRITERIA
- [ ] Every time the system is refreshed I want to have 10 default users configured per script
      - [ ] They should have normal realistic names and attributes filled out
           - First Name
           - Last Name
           - EMail 
           - Birthday
           - Telefon Number
           - Favorite Sports
           - If no big overhead a profile picture
(If to difficult to come up with people than use http://de.fakenamegenerator.com/gen-random-gr-gr.php)

nikriek:
Actual complexity of 3

Issue #149, opened 2017-11-26
Email Notification Event


PoJo93:
AS logged in User 
I WANT TO get a email one day before my game telling me when my game will start
TO be notified when I have games and make sure everybody attend the games
ACCEPTANCE CRITERIA

24 hours before one game starts in which I am enrolled I want to receive an email in the email from my account saying:
- Hallo FIRST NAME,
dein nächstes Spiel EVENT NAME startet morgen um EVENT TIME mit EVENT PARTICIPANTS. Ihr trefft euch am folgenden Ort: EVENT LOCATION
Viel Spaß beim Spiel!

PoJo93:
AS logged in User
I WANT TO get a email one day before my match telling me when my match will start
TO be notified when I have match and make sure everybody attend the match
ACCEPTANCE CRITERIA

24 hours before one match starts in which I am enrolled I want to receive an email in the email from my account saying:

Hallo FIRST NAME,
dein nächstes Spiel EVENT NAME startet morgen mit EVENT PARTICIPANTS. Ihr trefft euch am folgenden Ort: MATCH PLACE
Viel Spaß beim Spiel!

nikriek:
Actual complexity of 5

PoJo93:
Currently I cannot judge because I cannot see any match plan

PR #268, opened 2017-12-17
Generate 10 fake users on release. Ref #234

nikriek:
On hold as it requires #149 to be merged before.

Connects #234 

Only review: 
* lib/tasks/generate_fake_users.rake
* spec/tasks/generate_fake_users_spec.rb
* Procfile


lib/tasks/generate_fake_users.rake
```diff
@@ -0,0 +1,23 @@
+require 'faker'
+
+desc 'Generate 10 fake user account if less than 10 users'
+task generate_fake_users: :environment do
+  if User.count < 10
+    sports = %w(Fussball Tennis Volleyball Golf Tischtennis)
+
+    10.times do
+      sports.shuffle
+      password = Faker::Internet.password
```
jkbe: Do we want to be able to log in with the credentials of these users? If so, can we access the password somehow?

lib/tasks/generate_fake_users.rake
```diff
@@ -0,0 +1,23 @@
+require 'faker'
+
+desc 'Generate 10 fake user account if less than 10 users'
+task generate_fake_users: :environment do
+  if User.count < 10
```
jkbe: Maybe use the inverse condition instead and replace with a guard clause? That would avoid some of the indentation in subsequent lines.

spec/tasks/generate_fake_users_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  it 'preloads the Rails environment' do
+    expect(task.prerequisites).to include 'environment'
+  end
+  it 'runs gracefully with no subscribers' do
+    expect { task.execute }.not_to raise_error
+  end
+
+  context 'Less than 10 users' do
+    it 'should generate 10 fake users' do
+      expect { task.execute }.to change { User.count }.from(0).to(10)
```
jkbe, line 13: Don't we always create an admin in `seeds.rb` and therefore can't be sure we have 0 users?

lib/tasks/generate_fake_users.rake
nikriek: Nop, this was not a requirement.

spec/tasks/generate_fake_users_spec.rb
nikriek: Yeah, but the database is cleaned for each tests. Seeds.rb should not effect a test.

lib/tasks/generate_fake_users.rake
nikriek: You cannot use early return here because of the block syntax

lib/tasks/generate_fake_users.rake
nikriek: However, Jonas as PO mentioned some potential changes. lets see

lib/tasks/generate_fake_users.rake
nikriek: You can use the `next` keyword to jump to the next block. Will do that

spec/tasks/generate_fake_users_spec.rb
jkbe: Ok, I wasn't sure.

nikriek:
TODO: Use seed

Issue #169, opened 2017-11-28
Spielplan for Tournaments


MerlindlH:
**As a** tournament participant
**I want to** have a page with all currently played matches (Spiele)
**In order to** to have an overview of what happens in the tournament
**Acceptance Criteria:**
 - builds on #35 
 - [x] After the start of the Tournament, a Spielplan should be calculated (with random teampairing and single elimination)
 - [X] There should be different Spieltage which show the Spiele of the current Spieltag / Tournament Stage
 - [x] Each Spieltag should contain the list of current Spiele (in a Tournament each Round will be calculated according to the results of the previous Spieltag)
 - [x] Spieltage should have placeholderstrings if game information (who won games) is not available yet. the placeholderstrings should contain the needed games (e.g."Gewinner aus Achtelfinalspiel 1 gegen Gewinner aus Achtelfinalspiel 2") 
 -  [x] The last Spieltag should contain the matches for determining the first, second and third winner ("Finale" and "Spiel um Platz 3" if wanted)

prototypical demo:

![turnirspielplan](https://user-images.githubusercontent.com/8302493/33373858-07e8fd90-d504-11e7-97a4-a02417776728.png)

Markus-Brand:
Old Size: 20
New Size: 8

MerlindlH:
note: no additional deadlines between the tournament steps are needed (cf. [PO meeting protocol of the 5th of Jaunary](https://github.com/hpi-swt2/sport-portal/wiki/PO-Meeting-5th-January))

PR #275, opened 2017-12-21
Adjust Match - participants model

ErikBrendel:
With these changes, it should be possible for losers of a match to be implicitly participants of another match. This is needed for "Spiel um Platz 3" (see #169), as well as for other Tournament gameModes.


spec/models/match_result_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe MatchResult, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
ErikBrendel: will receive a test, which will also cover that one line here ( https://coveralls.io/builds/14762192/source?filename=app%2Fmodels%2Fmatch_result.rb ), but not today anymore...

Jeilef (commented):
My team and I havent had any contact with matches yet. You did some clean up so it seems fine to me.

db/migrate/20171221143530_create_match_results.rb
```diff
@@ -0,0 +1,10 @@
+class CreateMatchResults < ActiveRecord::Migration[5.1]
+  def change
+    create_table :match_results do |t|
+      t.integer :match_id
```
fpottbaecker: Maybe use [`references`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Table.html#method-i-references)

app/helpers/matches_helper.rb
```diff
@@ -1,13 +1,8 @@
 module MatchesHelper
   def participant_link(participant)
```
fpottbaecker: Maybe call it `link_to_participant`

app/helpers/matches_helper.rb
```diff
@@ -1,13 +1,8 @@
...
     if participant.present?
-      participant_name = participant.name
-      if participant.is_a? Match
-        link_name = I18n.t 'matches.winner_of_match', match: participant_name
-        link_to link_name, match_path(participant)
-      else
-        link_to participant_name, team_path(participant)
-      end
+      path = participant.is_a?(MatchResult) ? match_path(participant.match) : team_path(participant)
```
fpottbaecker: What if the outcome of the match is already known?

It looks like it will still link to the matches page, even though it should link  to the winning team (the displayed name will be the name of the winning team).

app/models/match.rb
```diff
@@ -17,23 +17,10 @@
...
 
 class Match < ApplicationRecord
-  has_many :home_matches, as: :team_home, class_name: 'Match'
-  has_many :away_matches, as: :team_away, class_name: 'Match'
-  def matches
-    home_matches.or away_matches
-  end
   belongs_to :team_home, polymorphic: true
   belongs_to :team_away, polymorphic: true
   belongs_to :event, dependent: :delete
```
fpottbaecker: Maybe add a `has_many :match_results` or something to have references both ways (and have deletion/updates cascade correctly).

app/models/match_result.rb
```diff
@@ -0,0 +1,41 @@
...
+    match.last_match_of team
+  end
+
+  def knows_advancing_participant
+    advancing_participant.present?
+  end
+
+  def advancing_participant
+    if winner_advances
+      match.winner
```
fpottbaecker, line 37: I fear this might result in very costly queries when viewing the page of a single match...

Can we do some caching, if we use the appropriate [active record callbacks](http://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks)?

app/models/match_result.rb
```diff
@@ -0,0 +1,41 @@
...
+# Table name: match_results
+#
+#  id               :integer          not null, primary key
+#  match            :integer          (points to a Match Object)
+#  winner_advances  :boolean
+#
+
+class MatchResult < ApplicationRecord
+  belongs_to :match
+
```
fpottbaecker: Since you will always want the relevant match of a MatchResult when working with that MatchResult, you should [eagerly load](http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations) the `:match` in the MatchResult [`default_scope`](http://guides.rubyonrails.org/active_record_querying.html#applying-a-default-scope).

app/views/events/_schedule_tournament.html.erb
```diff
@@ -28,8 +28,8 @@
 
     <tr>
       <td><%= match.place %> <%= match.index %></td>
-      <td><%= match.team_home.winner != nil ? link_to(match.team_home.name, team_path(match.team_home.winner)) : link_to('Gewinner ' + match.team_home.name, match_path(match.team_home)) %></td>
-      <td><%= match.team_away.winner != nil ? link_to(match.team_away.name, team_path(match.team_away.winner)) : link_to('Gewinner ' + match.team_away.name, match_path(match.team_away)) %></td>
+      <td><%= match.team_home.is_a?(Team) ? link_to(match.team_home.name, team_path(match.team_home)) : link_to('Gewinner ' + match.team_home.name, match_path(match.team_home.match)) %></td>
+      <td><%= match.team_away.is_a?(Team) ? link_to(match.team_away.name, team_path(match.team_away)) : link_to('Gewinner ' + match.team_away.name, match_path(match.team_away.match)) %></td>
```
fpottbaecker: You should move this to the events helper.

db/migrate/20171221143530_create_match_results.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.boolean :winner_advances
+
+      t.timestamps
```
fpottbaecker: Since there appears to be no way to change the timestamps of a matchresult, do we really need those, or can we just rely on the timestamps of the match?

fpottbaecker (changes requested):

app/models/match_result.rb
ErikBrendel: performance gets addressed when needed

fpottbaecker (approved):

Markus-Brand (approved):

PR #276, opened 2017-12-21
Bruteforce/145/gamedayduration for leagues

Askir:
Kinda work in progress right now. I'm not very happy with the view code yet. Still would love some input here :)
This is how the view looks right now:
gamedayduration

Not very pretty I know but it probably fullfills the acceptance cirteria :P


spec/factories/leagues.rb
```diff
@@ -4,6 +4,8 @@
     gameday_duration 7
 
     factory :league_with_teams, class: League, parent: :event_with_teams, traits: [:has_dates] do
+      game_mode :round_robin
```
usame-algan, line 7: Isn't it better to set the `game_mode` randomly here and specify it in the respective tests?

spec/models/league_spec.rb
```diff
@@ -21,14 +21,27 @@
...
-  it "is valid when produced by a factory" do
+  let(:league) { FactoryBot.build(:league) }
+  it 'is valid when produced by a factory' do
     league = FactoryBot.build(:league)
     expect(league).to be_valid
   end
 
-  describe "Generating league schedule with default values" do
+  it 'should not validate without gameday duration' do
+    league = FactoryBot.build(:league, gameday_duration: nil)
```
usame-algan: Is this any different than saying `league.gameday_duration = nil`? Since you create a league object already in line 26

spec/views/events/form_examples.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+# This set of shared examples provides a configurable way to test event creation and edit forms
+shared_examples 'an event form' do |for_class: Event, path: :new, with: []|
+
+  if path == :new
```
usame-algan: This is the same path as in the `elsif` statement. Do you mean `:edit`?

app/views/events/schedule.html.erb
```diff
@@ -21,7 +21,7 @@
         <tr class="gameday">
           <th colspan="6">
             <b>
-              <%= t '.gameday' %> <%= match.gameday %>
+              <%= t '.gameday' %> <%= match.gameday %> <%= "#{ (@event.date_for_gameday match.gameday).strftime("%d.%m.") } #{t('.to')} #{((@event.date_for_gameday match.gameday+1)-1.day).strftime("%d.%m.")}"%>
```
usame-algan: I think having the gameday and dates in one Tag is fine for now. We might have to seperate them later for more flexibility when styling them.

spec/views/events/form_examples.rb
Askir: yes, thank you ^^"

spec/models/league_spec.rb
Askir: no it's not I'll change so it's easier to read

app/views/events/schedule.html.erb
Askir: I didn't want to waste time styling when we have a whole team assigned to that :P

MarcelSchmidberger (changes requested):
This look pretty good, but I would suggest to add an Delimiter between the game day and the date, to visually separate the game day number and the date. Just like this: "Spieltag 1 - 06.01. bis 12.01." What do think about "-" as a Delimiter or would you suggest something else?
image

spec/factories/events.rb
```diff
@@ -45,6 +45,24 @@
       deadline { Date.current - 1 }
     end
 
+    trait :with_teams do
+      transient do
+        teams_count 5
```
laurenzse, line 50: Why do those values have to be fixed here? Can't we specify them randomly?

spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    assign(:matches, @league.matches)
+  end
+
+  it 'renders without errors' do
+    render
+  end
+
+  it 'has correct gameday date for gamedays' do
+    render
+    expect(rendered).to have_text '24.12. bis 30.12.'
```
laurenzse: Can you somehow internationalize this string?

spec/factories/events.rb
Askir: teams_count is just a default value here. You can override it by setting `teams_count: your number`

spec/views/events/schedule.html.erb_spec.rb
Askir: Yes, thanks I forgot about that.

spec/factories/leagues.rb
Askir: The value is defaulted to :round_robin right now because it is the only working game mode atm I think. If we change this we'd have to change it in the default factory aswell, which would cause a bigger changing of tests.
I want to start a test refactoring anyways since a lot of the older event tests seem a little of, so I'd add that to the ToDo list on that issue.

Also having subfactories for "league_with_teams" doesn't really make any sense which is why I added the trait. I don't want to remove this right here either though, because it would mean that I have to change another 10 or so files. And the PR is big enough as it is ^^"

Askir:
Added the delimiter to the page as you wished @MarcelSchmidberger .

MarcelSchmidberger (approved):
Thanks for adding the delimiter!

app/models/league.rb
```diff
@@ -56,4 +57,16 @@ def round_robin_pairings(teams_array)
   def can_join?(user)
     single? && (not has_participant?(user)) && (not deadline_has_passed?)
   end
+
+  def date_for_gameday(gameday)
+    ((gameday - 1) * gameday_duration).days.since startdate
```
erksch: Very sweet 👍

app/models/league.rb
```diff
@@ -56,4 +57,16 @@ def round_robin_pairings(teams_array)
...
+  end
+
+  def startdate_for_gameday(gameday)
+    date_for_gameday gameday
```
erksch: Why should this be neccessary?

app/views/events/_schedule_league.html.erb
```diff
@@ -4,7 +4,9 @@
     <tr class="gameday">
       <th colspan="6">
         <b>
-          <%= t '.gameday' %> <%= match.gameday + 1 %>
+          <%= t 'events.schedule.gameday' %> <%= match.gameday %></b>
+        <b>
+          <%= "- #{ (@event.startdate_for_gameday match.gameday).to_s(:day_and_month) } #{t 'events.schedule.to' } #{(@event.enddate_for_gameday match.gameday).to_s(:day_and_month)}"%>
```
erksch: I think seperating this in multiple lines would make it more readable.
I would suggest something like:
```
<b>
  - 
  <%= (@event.startdate_for_gameday match.gameday).to_s(:day_an_month) %>
  <%= t 'events.schedule.to' %>
  <%= (@event.enddate_for_gameday match.gameday).to_s(:day_and_month) %>
</b>
```

More statements but no unreadable long line of text concats.

app/views/events/_schedule_league.html.erb
```diff
@@ -4,7 +4,9 @@
     <tr class="gameday">
       <th colspan="6">
         <b>
-          <%= t '.gameday' %> <%= match.gameday + 1 %>
+          <%= t 'events.schedule.gameday' %> <%= match.gameday %></b>
```
erksch: Small thing: make a line break for the `</b>`.

app/models/league.rb
Askir: Just readability and it might need some logic later if you want to be able to have multiple gamedays at the same day (duration minimum is 1 day atm)

erksch: I would say alright but some would say do not write code until you need it (SWA).

Askir: I suppose I'll rename date_for_gameday to startdate_for_gameday then ^^

app/views/events/_schedule_league.html.erb
Askir: Still need to add a space so that it doesn't read like 24.12.to31.12. but 24.12. to 31.12.
but yes splitting is a good idea anyways :)

erksch (approved):

PR #285, opened 2018-01-05
Implementation of multiple results per match.

citron0xa9:


spec/models/game_result_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe GameResult, type: :model do
+  it "is valid when produced by a factory" do
+    result = FactoryBot.build(:game_result)
+    result.save
```
nikriek: You dont need to save the result. Make it a one line liek

```
expect(FactoryBot.build(:game_result)).not_to be_valid
```

spec/models/game_result_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    expect(result).to be_valid
+  end
+
+  it "fails validation with non-number scores" do
+    result = FactoryBot.build(:game_result, score_home: "2a")
```
nikriek: You might also check for the existence of a `score_home` in the errors dict after calling `is_valid?`

```ruby
r = FactoryBot.build(:game_result, score_home: "2a")
expect {r.is_valid?}.to change { r.errors }.from([]).to([score_home: [...]])
# OR
expect {r.is_valid?}.to change { r.errors[:score_home].count }.from(0).to(1)
```

spec/models/game_result_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    result.save
+
+    expect(result).not_to be_valid
+  end
+
+  it "is valid without any scores given" do
+    result = GameResult.new
+    match = FactoryBot.create(:match)
```
nikriek: just pass it directly in the constructor and check for validity

spec/factories/matches.rb
```diff
@@ -22,10 +22,15 @@
     association :team_away, factory: :team
     association :team_home, factory: :team
     association :event, factory: :event
-    score_home { rand(10..20) }
-    score_away { rand(1..9) }
     gameday 0
     points_away 3
     points_home 1
+
+    factory :match_with_results do
```
nikriek: Have a look at rspec traits

nikriek (approved):
Looks good. just minor stuff

spec/factories/matches.rb
citron0xa9: I think this is not doable because the callback `after(:create)` is used

Issue #125, opened 2017-11-25
Create rankings for event (League)


svitlanakalancha:
As Event participant I can **track the results** of my event in order to **evaluate** them.

**ACCEPTANCE CRITERIA**

_Frontend_

- [ ] Each event has separate ranking page connected to the event page (general ranking for all events is out of scope)
- [ ] Page includes the table with results. The table has following columns:
- Rang, Team, Spielanzahl, Gewonnen, Unentschieden, Verloren, Tore (geschossene/verlorene), Anzahl von Punkten
- [ ] Players or teams originate from participation list of the event

_Backend_

- [ ] Results of the played matches are automatically tracked and counted in the ranking table
- [ ] Results of the match are automatically translated into the points. Points assignment depends on match system. Typically it looks like this:
> the winner receives 3 points,
> even score - each team 1 point,
> looser - 0 points
- [ ] if 2 players/teams have equal amount of points by the end of event, the winner is determined by counting amount of goals. **NOTE:** works differently for "Schweizer System"
- [ ] The table should be automatically sorted by achieved points
- [ ] The winner is placed on the top of the table
- [ ] Ranking table can't be edited manually

**NOTE**
Consider different systems by leagues and tournaments

**OUT OF SCOPE**
Implementation of specific counting systems for different kinds of sport

<img width="548" alt="screen shot 2017-11-24 at 15 53 46" src="https://user-images.githubusercontent.com/27727717/33229635-7b1d465c-d1d2-11e7-9b47-d724e3292d8c.png">

<img width="404" alt="screen shot 2017-11-25 at 10 57 10" src="https://user-images.githubusercontent.com/27727717/33229637-82420bca-d1d2-11e7-9da4-7cb18d8d717b.png">

birneamstiel:
**Development Tasks**

- [ ] Create `GET :event_ranking` route to `routes.rb`
  - Add tests for route
- [ ] Create controller action `events_controller#ranking`
  - Check wether our event's participants are *teams* or *individual players* (following points assume we have teams, logic for individual players has to be adjusted)
  - Check wether our event's type is *league* or *tournament* (following points assume event type is league. Logic for tournament has to be adjusted, do tournaments even have rankings?)
  - Calculate necessary columns by iterating over `teams` and `matches` (something like in the pseudocode listing below)
  - Sort teams by points
    - Pay attention to ACs when two teams have the same amount of points
[...]

Issue #100, opened 2017-11-20
automatic point calculation for match


MarcelSchmidberger:
As a league participant,
I want to see the points each player of a match received, without entering them myself
So that correctness is ensured and effort removed.

**Acceptance Criteria**

- [x] The point rules are determined in the event creation(for win, lose, and draw). The winner of a game is the player with the most goals (basic rule for now, later on possibly with minimal goal difference ). The winner of a match is the player who won the most games. Points should be given accordingly and should be displayed in the last column.
- [x] The  Spielplan should calculate the points distribution depending on the outcome of the match. 

- [x] When the points are calculated I can see them in the boxes in which I can normally enter the points but still have the functionality to reedit them manually

PoJo93:
Nice work! Even though there is still one issue. When I put in the result the points are calculated. When I change the outcome of the game after that there is no recalculation. @nikriek

PoJo93:
Already good but points are displayed in the wrong order

nikriek:
Solved :)

PR #286, opened 2018-01-06
Create rankings for events

birneamstiel:
See #125


app/views/events/ranking.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <th><%= t '.lost_count' %></th>
+    <th><%= t '.goals' %></th>
+    <th><%= t '.goals_against' %></th>
+    <th><%= t '.goals_difference' %></th>
+    <th><%= t '.points' %></th>
+  </tr>
+  </thead>
+  <tbody>
+    <% @ranking_entries.each do |entry| %>
+    <tr>
```
birneamstiel, line 22: Whole block could be replaced by something like this:
``` html
<tr>
  <% entry.values do |value| %>
    <td><%= value %></td>
  <% end %>
</tr>
```
Not sure what is more readable @bakoe @jonasnoki

app/controllers/events_controller.rb
```diff
@@ -83,6 +88,80 @@ def schedule
     @schedule_type = @event.type.downcase!
   end
 
+  # GET /events/1/ranking
+  def ranking
```
birneamstiel, line 102: @bakoe @jonasnoki was sagt ihr zu dem jetzigen kontrollfluss? Mir erschien es am lesbarsten, den ranking_entry als parameter durchzureichen und dort die daten einzutragen...
Eine andere denkbare Möglichkeit wäre, recht viel der Logik dem Match zugeben. Dann hätte das Match Methoden wie `#get_ranking_entry_for_home_team` oder so... Was sagt ihr?

app/controllers/events_controller.rb
```diff
@@ -83,6 +88,80 @@ def schedule
...
+      ranking_entry.rank = index + 1
+    end
+  end
+
+  def parse_matches_data_into_ranking_entry(team, ranking_entry, matches, parse_match_details_for_home_or_away)
+    matches.each do |match|
+      score_home = match.score_home
+      score_away = match.score_away
+      match_has_score = score_home && score_away
+      next unless match_has_score
```
bakoe: Nice! 👍

app/views/events/ranking.html.erb
bakoe: Since the table's headings (lines 8 to 17 in the same file) cannot be localized dynamically from the Struct's values (or can they?), I would prefer the currently implemented solution of manually listing all of the Struct's properties in `<td>` tags. 

However, if there is an elegant solution of localizing the table's headings dynamically based on the Struct's properties/fields, we could replace both the actual value rendering (lines 23 to 32) **and** the table heading rendering (lines 8 to 17) with something like the snippet you proposed.

app/controllers/events_controller.rb
```diff
@@ -83,6 +88,80 @@ def schedule
...
+    # TODO Either add the possibility to calculate rankings for single-player events
+    # TODO or replace `teams` with the to be introduced EventParticipant superclass
+    # Leaves the Array of RankingEntry Structs empty when no teams participate in the event
+    @event.teams.each do |team|
+      ranking_entry = RankingEntry.new(nil, team.name, 0, 0, 0, 0, 0, 0, 0, 0)
+
+      event_matches = @event.matches
+      # Considers only the team's home matches that belong to the event
+      home_matches_in_event = team.home_matches & event_matches
+      parse_matches_data_into_ranking_entry team, ranking_entry, home_matches_in_event, :parse_match_details_for_home
```
bakoe, line 113: We could think about using destructuring (as described e. g. [here](http://tony.pitluga.com/2011/08/08/destructuring-with-ruby.html)) instead of passing the object to be worked on (the `ranking_entry`). The could could look like this:

```ruby
added_match_count, added_won_count, added_draw_count, 
added_lost_count, added_goals, added_goals_against, 
added_points = get_ranking_entry_additions_from_match team, 
                                                      home_matches_in_event, 
                                                      :get_ranking_entry_additions_for_home

ranking_entry.match_count += added_match_count
ranking_entry.won_count += added_won_count
ranking_entry.draw_count += added_draw_count
ranking_entry.lost_count += added_lost_count
ranking_entry.goals += added_goals
ranking_entry.goals_against += added_goals_against
ranking_entry.points += added_points
```

However, while writing this comment I noticed that this approach would result in way more bloated and less readable code. So we should stick with the way of passing the `ranking_entry` object around that you introduced! :+1:

bakoe (commented):
@birneamstiel I reflected on your approach regarding the passing of the `ranking_entry` object and the name-based method calling via `send` and came to the conclusion that the approach you chose seems very reasonable to me! I would not change the implementation of the EventsController's `ranking` method and its helper methods. 👍

app/controllers/events_controller.rb
bakoe: As mentioned in my PR feedback, I approve the current control flow. I would prefer to not move the helper methods into the `Match` class since this would require that one more component knows about the RankingEntry Struct and knows how to handle it properly. 

Maybe a future user story will require to re-use the RankingEntry-specific methods of the EventsController in some other controller – to my mind, this would be a good time to factor out the methods into the `Match` class. But as long as this encapsulation is not needed and the RankingEntry Structs get only used within the `Event` scope, I would leave the methods were they are currently.

app/controllers/events_controller.rb
```diff
@@ -93,6 +100,80 @@ def schedule
...
+    ranking_entry.goals_against += score_away
+    ranking_entry.points += match.points_home
+  end
+
+  def parse_match_details_for_away(ranking_entry, score_home, score_away, match)
+    ranking_entry.goals += score_away
+    ranking_entry.goals_against += score_home
+    ranking_entry.points += match.points_away
+  end
+
```
lheilmann, line 175: Great control flow and encapsulation of functionality in the methods above. :+1:

app/views/events/ranking.html.erb
lheilmann: I agree with @bakoe. In my opinion it is more readable, when both parts are either composed to a snippet or decomposed.

app/views/events/show.html.erb
```diff
@@ -98,6 +98,10 @@
             event_overview_path(@event),
             class: 'btn btn-default' %>
 
+<%= link_to t('.to_ranking'),
```
lheilmann, line 105: Please use consistent translation function call - either with or without brackets. See line 8 to 17 in same file.

spec/controllers/events_controller_spec.rb
```diff
@@ -362,6 +378,120 @@
...
+    end
+    describe 'when teams have joined the event' do
+      describe 'when no match has been played yet' do
+        it 'should calculate a lexicographically sorted zero-filled ranking' do
+          event = League.create! valid_league_attributes
+          team1 = FactoryBot.create(:team)
+          team2 = FactoryBot.create(:team)
+          event.add_team(team1)
+          event.add_team(team2)
+          event.generate_schedule
```
lheilmann: Why do you need to generate a schedule in `when no match has been played yet` but not in the following test cases in `when at least two matches have been played already`?

spec/controllers/events_controller_spec.rb
```diff
@@ -362,6 +378,120 @@
...
+        it 'should sum up the own goals of a participant correctly' do
+          expect(@ranking_entries.first.goals).to eq(20)
+        end
+        it 'should sum up the goals for the other side of a participant correctly' do
+          expect(@ranking_entries.first.goals_against).to eq(11)
+        end
+        it 'should calculate the number of points of a participant correctly' do
+          expect(@ranking_entries.first.points).to eq(4)
+        end
+        it 'should calculate the rank of two participants with the same points based on the number of goals correctly' do
```
lheilmann, line 460: Cool! As far as I have seen you covered all possible orders of RankingEntries here! :+1:

spec/controllers/events_controller_spec.rb
```diff
@@ -362,6 +378,120 @@
...
+          expect(@ranking_entries.first.name).to eq(@team1.name)
+          expect(@ranking_entries[1].name).to eq(@team3.name)
+
+        end
+      end
+    end
+    describe 'when users have joined the event' do
+      # TODO Factor out and reference tests for events with joined teams from above
```
lheilmann: I do not understand your `TODO` comments here. Are these test case moved to a different location or not implemented yet?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -50,6 +50,11 @@
       render
       expect(rendered).to_not have_selector(:link_or_button, t('helpers.links.destroy'))
     end
+
+    it 'has a ranking button' do
```
lheilmann, line 54: Nice!

spec/controllers/events_controller_spec.rb
```diff
@@ -53,6 +53,22 @@
     FactoryBot.build(:league, name: nil).attributes
   }
 
+  let(:valid_small_league_attributes) {
+    FactoryBot.build(:league, owner: @user, max_teams: 2).attributes
+  }
+
+  let(:valid_team1_attributes) {
+    FactoryBot.attributes_for(:team)
```
lheilmann: I assume the method `attributes_for` also creates an instance of Team by FactoryBot and then returns the attributes? Then `valid_team1_attributes`, `valid_team2_attributes` and `valid_team3_attributes` are instantiated with different values, which is probably the desired behaviour? I am asking because I have only used this method for returning attributes of an already created instance earlier.

lheilmann (changes requested):
Looks very good to me all in all! Especially control flow in EventsController is pretty cool and well-considered! :+1: I have only reviewed the code so far, because acceptance tests are not possible right now due to bugs in the GUI. Will do later, as soon as GUI is working properly again. Not our job...

lheilmann:
No need to fix all codeclimate issues in my opinion. Especially the issues targeting the control flow are not quite reasonable I think. So you can resolve some of them, as long as you have the same opinion. What about the `TODO`'s and `FIXME`'s?

app/controllers/events_controller.rb
```diff
@@ -93,6 +100,80 @@ def schedule
...
+      next unless match_has_score
+
+      ranking_entry.match_count += 1
+      parse_match_result_into_ranking_entry team, match, ranking_entry
+
+      send(parse_match_details_for_home_or_away, ranking_entry, score_home, score_away, match)
+    end
+  end
+
+  def parse_match_result_into_ranking_entry(team, match, ranking_entry)
```
Askir, line 152: All these methods don't really belong to the controller. Try to keep as much logic as possible outside of the controller. Methods that aren't actions need a good reason to even exist. And in this case it is very model specific. I would suggest encapsulation these methods into the Rankinglist model or even better into seperate plain old ruby objects (POROs) that handle all the neseccary logic.
The ranking function should do much more than setting `@ranking_entries = @event.ranking_entries`.

Askir: Here is a somewhat related article on the matter of POROs: http://blog.steveklabnik.com/posts/2011-09-06-the-secret-to-rails-oo-design

lheilmann:
Started with acceptance tests after design issues are fixed. Please see screenshot. Seems to me, that this is not the expected behaviour? Will continue with acceptance tests tomorrow in the morning.
screen shot 2018-01-17 at 00 12 37

bakoe:
@lheilmann I looked into the issue you documented and found out that it was probably caused by a bug/counter-intuitiveness in the “Spielplan” view:

The problem is: The “Spielplan” view currently makes the user think that the points entered in the left text field are the `points_home` (since the left team is `team_home` and the left side of the goals is `score_home`). **But:** The points entered in the left text field are actually `points_away` and the points entered in the right text field are actually `points_home`. See the following screenshot for explanation:

bildschirmfoto 2018-01-17 um 00 46 49 kopie

The corresponding `Match` object looks as follows, proving the documented bug/counter-intuitiveness:

bildschirmfoto 2018-01-17 um 00 47 37

As you can see, I as a user would expect the (left) `score_home` to be 3 and the (right) `score_away` to be 0. However, the opposite is the case.

Because of this bug, the “Spielplan” results from my screenshot result in the following ranking table which looks pretty similarly wrong to the one you posted @lheilmann:

bildschirmfoto 2018-01-17 um 00 47 46

However, the ranking table works actually as expected since the ranking entries (more precise: especially the amount of won, drawn and lost games as well as the amount of points) are **calculated based on the points** and _not based on the goals_. This is caused by the current match architecture of having both goals and (redundant) points – we decided to calculate the ranking entries based on the points since the points should theoretically get calculated from the goals automatically (see #100 with still open PR #292). The goals only get used for calculating the rank when two teams have the same amount of points.

TL;DR: Problem not caused by EventsController's `ranking` method but by a bug/counter-intuitiveness in the “Spielplan” view's text field ordering which causes the user to enter wrong points for matches. When #292 is merged (and thus #100 is closed), the problem should not occur anymore since the user should not have to enter points manually anymore.

lheilmann:
Thanks for the detailed explanation @bakoe! I would like to review this PR (acceptance tests) together with PO @svitlanakalancha in our meeting today. That's just because I think you need to know some tricky stuff or bugs, caused by other teams, to reproduce the desired behaviour. We then have maximum half an hour to review this issue all together. 

I approve this PR, because you covered all use case with test cases, so we can tell `ranking` is pretty much doing what it should be doing.

Please have a look at the codeclimate issues, so that we are able to merge this PR into `dev` before our meeting at 1 PM. 

Hang loose :+1:

lheilmann (approved):

app/controllers/events_controller.rb
birneamstiel: Thanks! Refactoring will follow up:  https://github.com/hpi-swt2/sport-portal/issues/367

PR #291, opened 2018-01-06
Kickerz/#219/remove team ids in view

niconomaa:
ensure that teams or always returned ordered by their creation date and remove unintuitive ids from team index view


spec/models/team_spec.rb
```diff
@@ -68,4 +68,10 @@
     expect(team.in_event?).to be true
   end
 
+  it "by default return teams ordered by their date of creation" do
+    team = FactoryBot.create :team
+    another_team = FactoryBot.create :team
+    team.name = "New Name"
```
bakoe: Maybe consider adding a comment to this line explaining why it is necessary: to check that the teams are not ordered by `updated_at`. At first glance, it was not clear to me why the line is necessary.

spec/features/team/index_team_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe "Index team page", type: :feature do
+
+  it "should render without an error" do
+    visit teams_path
+  end
+
+  it "should not contain an id column" do
+    @team = FactoryBot.create :team
+    visit teams_path
+    expect(page).to_not have_text "Id"
```
bakoe: Maybe consider using `Team.human_attribute_name :id` instead of the hardcoded “Id” string here.

app/views/teams/index.html.erb
```diff
@@ -14,7 +13,6 @@
   <tbody>
     <% @teams.each do |team| %>
       <tr>
-        <td><%= link_to team.id, team_path(team) %></td>
```
bakoe, line 17: Since the table's “id” column contained links to the specific team pages before the column got removed, we may consider adding the link to the “name” column instead. Maybe a question for our PO @svitlanakalancha?

bakoe (approved):
Some minor comments – apart from that, looks very good to me! 

Have tested in browser and can confirm that the removal of the column as well as the sorting by creation date work as specified.

lheilmann (approved):
Looks perfect to me!

Issue #197, opened 2017-12-02
Configure Matches


PoJo93:
AS the organizer of an event 
I WANT TO configure how the games are played saying that I want to set the game mode, determine when someone wins and configure how many points the winner, as well as the looser, earns
TO control how each match will be played

ACCEPTANCE CRITERIA
- [ ] When creating and event or when editing an event as an organizer I want to have additional options with already set default values
  - [ ] When I click on the button "Erweiterte Einstellungen" I want to see additional configuration options for all matches
![additional_setting_sportportal](https://user-images.githubusercontent.com/5064619/33808316-9bd51f46-dde4-11e7-9a89-d03f4b9cc335.png)
![image 2017-12-10 20 31 35](https://user-images.githubusercontent.com/5064619/33808562-1f3d4b02-dde9-11e7-9602-229ebef97699.jpg)

      - [ ] When I click on the drop-down menu next to the label "Spielart:   Best of" there are the options 1-8 whereas 1 should be the default. When I hover over the label you can see "(z.B. Fussball = Best-of-1, Tennis = Best-of-3)"
      - [ ] When I click on the drop-down menu next to "Gewinnregel je Spiel" there is one option which says "Sieger ist, wer die meisten Sätze gewinnt"
      - [ ] Next to the label "Punkteverteilung" there are the labels 
          - "Gewinner" and a box in which I can only enter a positive Integer Number (default 3)
          - "Unentschieden" and a box in which I can only enter a positive Integer Number (default 1)
          - "Verlierer" and a box in which I can only enter a positive Integer Number (default 0)

- [ ] When I am in editing mode of an event there is a button "Speichern". When I am in creating an event there stays the Button "Veranstaltung erstellen. When clicking the button it should
            - [ ] store what I just put on the formula or when I didn't put anything on the advanced options store the default values
            - [ ] and set the options for every match in the event (also already created matches)
- [ ] Update the hardcoded points calculation in the match model

nikriek:
Please modify the function after https://github.com/hpi-swt2/sport-portal/pull/292 is merged to reflect the different game modes in the point calculation

PoJo93:
@citron0xa9 I don't see a Field for the Best of:  option

citron0xa9:
@PoJo93 should be fixed now

PoJo93:
@citron0xa9 ok nice. According to the User story there are three points missing

[...]

PR #292, opened 2018-01-07
Add automatic point calculation for match

nikriek:
connects #100


nikriek:
connects #100

spec/models/match_spec.rb
```diff
@@ -35,4 +35,32 @@
...
+      let(:match) { FactoryBot.build(:match, :empty_points, score_home: 0, score_away: 1) }
+      it { is_expected.to change { match.points_home }.from(nil).to(0) }
+      it { is_expected.to change { match.points_away }.from(nil).to(3)  }
+    end
+
+    context 'team home and team away have an equal score' do
+      let(:match) { FactoryBot.build(:match, :empty_points, score_home: 1, score_away: 1) }
+      it { is_expected.to change { match.points_home }.from(nil).to(1) }
+      it { is_expected.to change { match.points_away }.from(nil).to(1) }
+    end
```
citron0xa9, line 61: case for "manually entered points not modified" could be added

citron0xa9 (approved):
Looks good

nikriek:
Yeah true. But it’s not part of the model 😉

nikriek:
@citron0xa9 Who merges first?

spec/models/match_spec.rb
nikriek: That is not part of the model

app/models/match.rb
```diff
@@ -81,4 +91,19 @@ def adjust_index_by(offset)
...
     save!
   end
+
+  def calculate_points
+    return if !has_scores? || has_points?
+
+    if score_home > score_away
+      self.points_home = 3
+      self.points_away = 0
+    elsif score_home < score_away
```
manner: Shouldn't the win, lose and draw points be retrieved from the corresponding event? Or is this only possible after #197 is merged?

manner (approved):

app/models/match.rb
```diff
@@ -92,18 +92,20 @@ def adjust_index_by(offset)
...
-      self.points_away = 0
+      set_points(3, 0)
     elsif score_home < score_away
-      self.points_home = 0
-      self.points_away = 3
+      set_points(0, 3)
     else
-      self.points_home = 1
-      self.points_away = 1
+      set_points(3, 3)
```
manner: should be `set_points(1, 1)` ;)

nikriek: Yeah, that was a copy-paste error. Good that we have tests ;)

app/models/match.rb
nikriek: yes and yes. But I can leave the hardcoded for now and you insert it ;)

PR #293, opened 2018-01-07
Kickerz/#247/introduce team member numbering

niconomaa:
replace id col by number col in team show page, ensure consistent ordering of team members by team-join-date


bakoe (commented):
Do you really want to remove the schema.rb file? To my mind, this file is rather important and should remain in the repository. 

👍 LGTM code-wise; haven’t tested yet though.

db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
```
birneamstiel: I don't think we want to have this in here :-)

spec/features/team/show_team_spec.rb
```diff
@@ -21,4 +21,22 @@
       expect(page).to have_text another_user.first_name
     end
   end
+
+  describe 'display consecutive numbering of members' do
```
birneamstiel: Actually, the consecutive numbering isn't tested here, but only the table head, right? It may be nice to test that two consecutively added members have consecutive numbers.

birneamstiel (changes requested):

app/views/teams/show.html.erb
```diff
@@ -35,17 +35,17 @@
       <thead>
           <tr>
             <th><%= t '.select', :default => t("helpers.select.prompt") %></th>
-            <th><%= model_class.human_attribute_name(:id) %></th>
+            <th>#</th>
```
lheilmann: Consider using a translation string here as well be stay consistent. You would have to adjust a test case as well.

spec/features/team/show_team_spec.rb
```diff
@@ -21,4 +21,22 @@
...
+    before(:each) do
+      sign_in @user
+    end
+
+    it 'should not show an id column for members' do
+      @team.members << @user
+      visit team_path @team
+      expect(page).to_not have_text "Id"
```
lheilmann: You maybe want to test the expected test with a translation string and no hardcoded string.

spec/features/team/show_team_spec.rb
```diff
@@ -21,4 +21,22 @@
...
+    end
+
+    it 'should show # column for members' do
+      @team.members << @user
+      visit team_path @team
+      expect(page).to have_text "#"
```
lheilmann: See earlier comment about using translation strings.

lheilmann (changes requested):

db/schema.rb
niconomaa: i think we should consider keeping it! let's talk about this on Wednesday

spec/features/team/show_team_spec.rb
niconomaa: https://www.youtube.com/watch?v=2nsbDyjn3L4

birneamstiel (approved):

lheilmann (approved):

PR #304, opened 2018-01-09
Kickerz/#249/render leave team button only for members

lheilmann:


bakoe (approved):

bakoe:
I have manually tested this in the browser and can confirm it's working. :-)

PR #313, opened 2018-01-12
Issuenumber5/#223/remove players page

pawelsoundso:
Remove the "players" button in the navigation bar and restrict the access to the user-index page for normal users.


Jeilef (approved):
well done

PR #314, opened 2018-01-12
Issuenumber5/#221/min max players in teams

fswt:


app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -58,5 +59,29 @@ $( document ).on('turbolinks:load', function() {
            $("#event_enddate").val(formattedDate);
        }
     });
+
+    // Autofill of player count for an event
+    $("#event_player_type").on("change", showPlayerCount);
+    $("#event_min_players_per_team").hide();
```
pawelsoundso: Why do you set the player count here too? It is set in line 64 anyways... This might cause the error you were having with the fields not showing up anymore if you reload the site (because you hide it at the beginning anyways at every time).

app/controllers/events_controller.rb
```diff
@@ -102,6 +103,13 @@ def set_event
       @event = Event.find(params[:id])
     end
 
+    def set_playercount_per_team
+      if event_params[:player_type] == "single" || @event.type == 'Rankinglist'
```
pawelsoundso: You could move this to set_associations, since it would make more sense to be integrated there (in the end this is an association as well). Also you could use `@event.player_type == Event.player_types[:single]` there instead of the `event_params` stuff. Or you get rid of this entirely and set a default value in the database for `min_players_per_team` and `max_players_per_team`, which would simplify some other things like the factories as well.

app/models/event.rb
```diff
@@ -125,6 +125,17 @@ def human_game_mode
     self.class.human_game_mode game_mode
   end
 
+  def fitting_teams(user)
+    fitting_teams = user.owned_teams.multiplayer
+    fitting_teams.each do |team|
+      team_member_count = team.members.count
+      if not (min_players_per_team <= team_member_count && max_players_per_team >= team_member_count)
```
pawelsoundso: Move this into a separate method like `is_eligible?(team)`,  where you set line 131 and return the condition of 132 (and let's negate the condition so we can get rid of the `not`) 😄.

spec/models/event_spec.rb
```diff
@@ -104,4 +104,54 @@
...
+    end
+
+    it "should be possible to join a team event if min/max players per team fits to team size" do
+      @team_event.min_players_per_team = 6
+      @team_event.max_players_per_team = 6
+
+      expect(@team_event.fitting_teams(@user).count).to be(1)
+    end
+
+    it "should be possible to join a team event if min/max players per team fits to team size" do
```
pawelsoundso: You should change the names of these tests according to what they're doing.

pawelsoundso (changes requested):
Only some minor adjustments are needed here. Otherwise it's well done!

app/controllers/events_controller.rb
```diff
@@ -102,6 +103,13 @@ def set_event
...
+        @event.min_players_per_team = 1
+        @event.max_players_per_team = 1
```
sumpfranze: is this reasonable? Should the feature be disabled or is it not possible that there will be ranking lists for teams as well? If, at some point (or even already now), it should be possible that teams participate in a ranking list, this needs to be changed.

app/views/events/forms/_Tournament.html.erb
```diff
@@ -1,6 +1,20 @@
...
+    <%= form.number_field :min_players_per_team, class: 'form-control', min: 2, :value => 2 %>
+  </div>
+</div>
+
+<div class="form-group" id="event_max_players_per_team">
+  <%= form.label :max_players_per_team, class: 'control-label col-lg-2' %>
+  <div class="col-lg-10">
+    <%= form.number_field :max_players_per_team, class: 'form-control', min: 2, :value => 2 %>
+  </div>
+</div>
```
sumpfranze: see _League.html.erb

app/views/events/forms/_League.html.erb
```diff
@@ -1,6 +1,20 @@
...
+    <%= form.number_field :min_players_per_team, class: 'form-control', min: 2, :value => 2 %>
+  </div>
+</div>
+
+<div class="form-group" id="event_max_players_per_team">
+  <%= form.label :max_players_per_team, class: 'control-label col-lg-2' %>
+  <div class="col-lg-10">
+    <%= form.number_field :max_players_per_team, class: 'form-control', min: 2, :value => 2 %>
+  </div>
+</div>
```
sumpfranze: if this should only be possible for Leagues and Tournaments, I would put it in a separate partial form, since this would reduce code duplication.

Do it analogue to dates and player_type.

config/locales/de.yml
```diff
@@ -95,6 +95,8 @@ de:
         duration: "Dauer in Tagen"
         metric: "Metrik"
         initial_value: "Anfangswert"
+        min_players_per_team: "Min. Anzahl Spieler pro Team"
```
sumpfranze: maybe call this "minimale Teamgröße" and "maximale Teamgröße"? sounds a bit more natural

app/controllers/events_controller.rb
fswt: As far as I know, ranking lists are single-player-only.

app/models/event.rb
```diff
@@ -125,6 +125,17 @@ def human_game_mode
...
+        fitting_teams.delete(team)
```
fswt: Bug: It seems that through this line the team is also deleted in the DB.

pawelsoundso (approved):

PR #315, opened 2018-01-13
Flexibus/#185 2/sign in and sign up design

### LINKED ISSUE #185: Sign-in and Sign-up design
Design von Seiten "Anmelden" und "Registrieren"
- all data (for example font size, colors etc - in mock, ask Dima for CSS-Code);
- FONT: ProximaNova
- footer is implemented separately in #186 !

**Estimation: XL**

 
 
Placeholder shouldn't disappear but move upwards (if not doable, ask Dima for new design without transitions), remove link 'Registrieren'

 
 
Both should be mobile-friendly: (navbar has background image)

 

- [x] Research on library for transitions on sign up page
- [x] Registrieren page
- [x] Anmelden page

jcremerius:
Closes #185.


app/views/application/_background.html.erb
```diff
@@ -0,0 +1,37 @@
+<%-
+  $background = "navbar-overlap"
+  $new_background = "navbar-overlap background"
+  $no_background = "navbar-overlap"
+  case (request.user_agent =~ /Mobile|webOS/).present? # check if user is using a mobile device
+    when true   # in case it is a mobile device
+      case controller_name
+        when ""
```
emanuel-metzenthin: Both conditional are equal. 'Case' can be left out.

app/views/layouts/application.html.erb
```diff
@@ -6,17 +6,26 @@
     <title><%= t('appname') %></title>
     <%= csrf_meta_tags %>
 
-    <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
-    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <% if (action_name === "new" && controller_name == "sessions" or controller_name === "users" && action_name == "new" or controller_name === "users" && action_name == "create") %>
+        <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track'=>'reload' %>
+        <%= stylesheet_link_tag 'navbar', media: 'all', 'data-turbolinks-track'=>'reload' %>
```
emanuel-metzenthin: navbar.css tag could be put in navbar.html.erb and this condition left out.

app/views/layouts/application.html.erb
```diff
@@ -6,17 +6,26 @@
...
+        <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track'=>'reload' %>
+    <% end %>
+
+
+    <%= javascript_include_tag 'application', 'data-turbolinks-track'=> 'reload' %>
   </head>
 
-  <body>
+  <!--Display the background -->
+  <body class="<%= render partial: "application/background"%>">
```
emanuel-metzenthin, line 19: Just a thought: Wouldn't it be more elegant to put the computation of the background classes in a helper method inside of the application controller? Similar to the error_detector method.

emanuel-metzenthin (changes requested):

app/views/application/_background.html.erb
MariusDanner: Done

app/views/layouts/application.html.erb
MariusDanner: But then the style link would be inside the body, which is at least bad style.
But we can move the application tag outside of the conditional

MariusDanner (approved):
Looks good to me, but i would not trust me cause i wrote some of that :D

jkbe:
@emanuel-metzenthin please recheck.

app/views/application/_background.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        when "sessions"
+          case action_name
+            when "new"
+              $background = $new_background
+            else
+              $background = $no_background
+          end
+        when "users"
+          case action_name
+            when "new"
```
deeps96, line 19: Since the actions new and create do have the same behaviour, you can write something like `when "new", "create"` instead of having multiple "when".

app/views/application/_background.html.erb
```diff
@@ -0,0 +1,37 @@
+<%-
+  $background = "navbar-overlap"
+  $new_background = "navbar-overlap background"
+  $no_background = "navbar-overlap"
```
deeps96, line 4: Whats the purpose of background having the content of no_background already and assigning it again? As it seems to have no background per default, it´s not necessary to set the "no-background" again. Maybe it would be a good idea to split this method into submethods (mobile desktop) to increase the readability.

deeps96 (commented):
Looks good! I just have some few suggestions for further improvement.

app/assets/stylesheets/sign_in.css
```diff
@@ -0,0 +1,116 @@
...
+/* Mobile/desktop dependent: */
+
+@media screen and (min-width: 480px) {
+    .background{
+        background: url(https://owncloud.hpi.de/index.php/s/siSJzCTBN5U3pbf/download) no-repeat center center fixed;
+        -webkit-background-size: cover;
+        -moz-background-size: cover;
+        -o-background-size: cover;
+        background-size: cover;
+        background-repeat: none;
```
deeps96, line 98: The value of background-repeat got set already in line 81.
Additionally none is not a valid value for the background-repeat property. See https://www.w3schools.com/cssref/pr_background-repeat.asp supported property values.

emanuel-metzenthin (approved):

jkbe:
@deeps96 as you're working on similar things on your branch, maybe you could fix the code smells mentioned by you there?

app/assets/stylesheets/sign_in.css
jkbe: Fixed this in #355.

PR #320, opened 2018-01-14
Swteam/#243/open id no password confirmation

DanielStephan97:


fpottbaecker (approved):

features/step_definitions/user.rb
```diff
@@ -140,3 +140,7 @@
 And(/^he logs out$/) do
   sign_out
 end
+
+And (/^he changes his email/) do
+  fill_in :user_email, with: 'test2@test.com'
```
Markus-Brand: This assumes, that the mail was not test2@test.com before...
Is changing this to 'new' + [old mail] feasable

Issue #124, opened 2017-11-25
Send email to team members


svitlanakalancha:
As Team Member I can send **emails** to other team members.

**Acceptance criteria:**

- [x] Every team member has rights to send message to other team members
- [x] Team member can choose group of players to send group email
- [x] As recipients are chosen and the button is clicked, the window of mailing program opens
- [x] If user has no mailing program, user can send emails via owa/etc. Recipients'adresses are shown in To: field

bakoe:
@svitlanakalancha:

While working on this issue, I noticed that the ”Bitte wähle“ and ”Aktionen“ columns as well as the “Wende folgende Aktion auf Ausgewählte an:” row are displayed even if the user currently logged **cannot** perform any action on any team member (because he is not a team member). See the following screenshot for explanation:

<img width="1233" alt="bildschirmfoto 2018-01-14 um 13 18 16" src="https://user-images.githubusercontent.com/807665/34915861-e04fac4a-f92d-11e7-9da2-a5dc69fce086.png">

I changed the view to only show the columns and rows mentioned when the user currently logged in **can** perform any action on any team member. See the following screenshot for explanation:

<img width="1247" alt="bildschirmfoto 2018-01-14 um 13 19 35" src="https://user-images.githubusercontent.com/807665/34915866-fe145b9a-f92d-11e7-951d-4fbc56c73953.png">

Now my question: Is it okay for you to have this ”bug“ (if it even was one?) fixed within this issue or do you want to have a separate issue for this bug? If I should revert the changes regarding the rendering of these columns and rows, it would be a matter of a minute – so it would be no problem for me development-wise to revert the changes I made.

svitlanakalancha:
@bakoe as discussed, it is ok. Thank you!

bakoe:
Actual time I invested in this issue: 1 h

PR #321, opened 2018-01-14
Allow sending emails to single and multiple team members from the show team page

bakoe:
**Note:** Within this issue, I changed the ”Bitte wähle“ and ”Aktionen“ columns as well as the “Wende folgende Aktion auf Ausgewählte an:” row to only be displayed/rendered when the user currently logged-in can perform any action on any team member (i. e. if he is a member of the specific team himself). See my comments regarding this included bugfix in #124. @svitlanakalancha approved making these changes within this issue via Slack (see channel `#124-emails-to-tmember`).


db/schema.rb
```diff
@@ -88,6 +88,7 @@
     t.text "description"
     t.string "kind_of_sport"
     t.boolean "private"
+    t.boolean "single", default: false
```
birneamstiel: I'm not sure what this is about, is it correct to have this included here?

birneamstiel (approved):
LGTM!

lheilmann (approved):
Tested completely in browser (Chrome). Acceptance tests and code looks very good to me!

db/schema.rb
bakoe: This seems to have resolved itself when the `dev` branch got merged into my feature branch (06385f7). 

Probably nobody commited the applied database changes (`schema.db`) – introduced by the newest migration – when I branched my feature branch off the `dev`. Since then, some other PR has probably commited the `schema.db` and been merged into the `dev`, thus removing my unintentional changes to the `schema.db` file from this PR's diff. 

[Since the `schema.db` file should always represent the database schema in its most recent version](https://stackoverflow.com/a/6450396), I think it's okay to include changes to the `schema.db` file even if the corresponding user story does not touch the database structure. Maybe we can discuss this further if a future PR does touch the `schema.db` again.

PR #327, opened 2018-01-14
Have a template for every user story

PoJo93:
Like this you don't have to retype everything and gain the same format among all user stories


ISSUE_TEMPLATE
```diff
@@ -0,0 +1,6 @@
+**AS** a user
```
fpottbaecker, line 1: This file should be in a `.github` folder to avoid cluttering. See [this](https://github.com/blog/2111-issue-and-pull-request-templates)

fpottbaecker (changes requested):

conradlempert (approved):

PR #347, opened 2018-01-15
Kickerz/#253/filter teams by membership

niconomaa:
add filtering functionality to team index view


app/controllers/teams_controller.rb
```diff
@@ -5,7 +5,12 @@ class TeamsController < ApplicationController
 
   # GET /teams
   def index
-    @teams = Team.multiplayer
+    @multi_teams = Team.multiplayer
+    if params[:filter] == I18n.t("helpers.teams.show_mine")
```
lheilmann: Using a translation string to indicate whether the filter flag is set or not is in my opinion not the best solution. Consider using a boolean or an integer here. For example `filter` could be set to `true`, if teams should be filtered to only see my teams. This also works better as an URL parameter. You would have to adjust `teams.js` and one of the test cases as well.

spec/features/team/index_team_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  it "should render without an error" do
+    visit teams_path
+  end
+
+  it "should have a select box to filter teams" do
+    visit teams_path
+    expect(page).to have_select('filter')
+  end
+
+  it "should get " do
```
lheilmann: Probably a description missing?

lheilmann (changes requested):
Looks good to me in general! Only one conceptual change request.

spec/features/team/index_team_spec.rb
niconomaa: never!!

lheilmann (approved):
Looks great man!

app/controllers/teams_controller.rb
niconomaa: yuup, changed it; but i initally decided against doing that for a couple of reason:

* since the param is passed from the view, it arrives in the controller as a string no matter what, imo. we could parse it in the controller, but we are nonetheless dealing with strings here...
* passing a boolean precludes us from extending our dropdown filter, since it is now binary. i thought that, because we are using a dropdown (and not a checkbox, for instace), possible extensibility of the filter might have been desired

i'd go with this solution anyways, though, dunno what's better
but looks much cleaner the new way

lheilmann: Alright! I can understand your thoughts on extensibility - we should think about this definitely again, when there is the requirement for multiple filtering options.

There is another reason, why you should not use translations here, besides the fact, that we are handling strings anyway. Think about the case when both translations you are using return the exact same string (by accident or whatever).

Issue #342, opened 2018-01-15
Multiple playing days per day


MarcelSchmidberger:
As organizer of a league
I want to be able to have many games per day
I order to organize one-day(or two day) events

**Acceptance Criteria**
- [ ] multiple game days per day

**To be clarified**
right now it is possible to enter 0 for the gameday length and schedule all game days on one day
we need to clarify whether that is sufficient for our customer.
If there is the possibility for manual changes of the dates it would be possible to change all game days, as you wish.

MarcelSchmidberger:
Manual changes were ok, as a measure to enable multiple gameday per day. This will be resolved #435

PR #354, opened 2018-01-15
Bruteforce/145/gamedayduration for leagues *corrections*

Askir:


app/models/league.rb
```diff
@@ -21,7 +21,7 @@
 
 class League < Event
   validates :deadline, :startdate, :enddate, presence: true
-  validates :gameday_duration, presence: true
+  validates :gameday_duration, presence: true, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 1000 }
```
usame-algan, line 24: Should it be possible to have a gameday duration of 0 or are we starting at 1?

usame-algan (changes requested):

app/models/league.rb
Askir: 0 is intended to enable the workaround mentioned in: #342

usame-algan (approved):

PR #360, opened 2018-01-16
Swteam/#245/unify profile

ErikBrendel:
We merged the `edit_profile` and `edit` - pages for Users. There is now one single settings page. In addition to that, the Profile page (`user::show`) displays more information. 3 example user accounts demonstrate how it will look.

We disagree with the three codeclimate issues and hope that you will also see these issues as "invalid"- or "wontfix" - candidates

The coverage decrease seems to be due to a reduction in coverable lines


app/controllers/users_controller.rb
```diff
@@ -1,7 +1,7 @@
 class UsersController < Devise::RegistrationsController
   # https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions
   helper_method :error_detector
-  load_and_authorize_resource :user, only: [:index, :show, :edit, :destroy, :confirm_destroy, :dashboard]
```
fpottbaecker: Show is no longer authorized?

app/views/users/edit.html.erb
```diff
@@ -28,6 +26,48 @@
...
     </div>
   </div>
+  <div class="form-group row">
+    <%= f.label :avatar, t('activerecord.models.image.one'), class: 'control-label col-lg-2'%>
+    <div class="col-lg-10">
+      <%= image_tag @user.avatar_url || image_path('missing_avatar.png'), id: 'preview', style: 'max-width:100px; max-height:100px'%>
+      <%= f.file_field :avatar, class: 'form-control-file' %>
+      <% if current_user.avatar %>
+        <%= f.check_box :remove_avatar %>
+        <label for="user_remove_avatar">
```
fpottbaecker: Use `f.label` here...

fpottbaecker (changes requested):

app/controllers/users_controller.rb
```diff
@@ -107,17 +92,17 @@ def self.unimportant_changes?(resource, params)
...
     # Overridden methods of `Devise::RegistrationsController` to permit additional model params
     def sign_up_params
       generate_random_password if get_omniauth_data
-      params.require(:user).permit(:first_name, :last_name, :email, :password, :image, :remove_image, :password_confirmation, event_ids: [])
+      params.require(:user).permit(:first_name, :last_name, :email, :password, :image, :remove_image, :password_confirmation, :avatar, :remove_avatar, :birthday, :telephone_number, :telegram_username, :favourite_sports, event_ids: [])
     end
 
     def account_update_params
-      params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation, :current_password, event_ids: [])
+      params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation, :current_password, :avatar, :remove_avatar, :birthday, :telephone_number, :telegram_username, :favourite_sports, event_ids: [])
```
nikriek, line 106: Would it make sense to split up normal profile and password updates? otherwise one would always have to confirm a profile update with a password

app/views/devise/registrations/show.html.erb
```diff
@@ -1,13 +1,22 @@
 <%- model_class = User -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => model_class.model_name.human.titleize %></h1>
+  <h1><%=@user.name %></h1>
 </div>
 
+<br>
+<div style="text-align: center;">
```
nikriek: Bootstrap has a class for this

db/seeds.rb
```diff
@@ -18,3 +18,41 @@
 admin.save(validate: false)
 
 FactoryBot.create(:match)
+
+# Some example user entries
+
+FactoryBot.build(:user,
```
nikriek: Please do not use factorybot. It should be used for creating test factories... Just use `User.create`

nikriek (changes requested):

db/seeds.rb
nikriek: https://robots.thoughtbot.com/factory_girl-for-seed-data

app/controllers/users_controller.rb
ErikBrendel: yep. Other people want to see your profile

app/controllers/users_controller.rb
ErikBrendel: there was a userstory that made a password confirmation only necessary for important changes. This feature already works, no worries :)

fpottbaecker (approved):

nikriek (approved):

PR #361, opened 2018-01-16
Swteam/#169: Third place Match

ErikBrendel:
With this PR, a **match for place 3** gets created for every Tournament.

We disagree with the two codeclimate issues and hope that you will also see them as "wontfix/invalid" - candidates


app/models/tournament.rb
```diff
@@ -42,10 +44,21 @@ def finale
     nil
   end
 
+  def place_3_match
+    matches.each do |match|
```
fpottbaecker: `matches.find`

app/models/tournament.rb
```diff
@@ -28,12 +28,14 @@ class Tournament < Event
...
+
+    special_standing = special_standing_of team
+    return special_standing if special_standing.present?
+
+    final_match.last_match_of(team).standing_string_of team
   end
 
   def finale
-    #TODO: keep care of 3rd place match here
     matches.each do |match|
```
fpottbaecker: `matches.find`

fpottbaecker (changes requested):

app/models/tournament.rb
```diff
@@ -36,32 +36,23 @@ def standing_of(team)
   end
 
   def finale
-    matches.each do |match|
-      if match.depth == 0
-        return match
-      end
-    end
-    nil
+    matches.where(gameday: finale_gameday).first
```
fpottbaecker: You can use `find_by`

fpottbaecker (approved):

ErikBrendel:
closes #169

Issue #316, opened 2018-01-13
Highlight current day in date picker


GittiHab:
**AS AN** editor of an event
**I WANT TO** the current day to be highlighted in the calendars
**SO THAT** I can easily find the current day

## ACCEPTANCE CRITERIA
- [x] When a date picker is shown, for example when setting the deadline or start date, the current day is highlighted

GittiHab:
Works

PR #366, opened 2018-01-17
fix the datepicker Ref.: #316

ConstantinLange:
current date is always highlighted.


Jeilef (approved):
I like your bootstrap

Issue #318, opened 2018-01-13
No deadlines in the past


GittiHab:
**AS AN** organiser of an event
**I WANT TO** to only select deadlines in the future or today
**SO THAT** I cannot create events nobody can join

## ACCEPTANCE CRITERIA
- [ ] Only dates in the present or future should be valid inputs for the "deadline" field when editing/creating events

GittiHab:
Works

PR #368, opened 2018-01-17
fix the datepicker Ref.: #318

ConstantinLange:
You could only change dates of the datepicker to future or present


Jeilef (approved):
good good

PR #382, opened 2018-01-17
Swteam/#168/place 3 optional

### LINKED ISSUE #168: create Tournament additions
**As a** Tournament organizer
**I want to** get additional information about the event depending on my inputs during Event creation
**In order to** plan my tournament better
**Acceptance Criteria:**
 - builds on #104
 - [ ] If max Teamzahl is given, benötigte Spieltage should be calculated
 - [ ] if benötigte Spieltage is known,  and Startdatum and Länge eines Spieltags in Tagen are entered, the estimated Enddatum should be shown (and be refreshed) accordingly.
 - [ ] add a checkbox to disable "Spiel um Platz 3"

Markus-Brand:
With this PR, the Match for place 3 in tournaments got optional: While creating a new tournament, you can select whether there should be one or not. When changing this setting on an already created tournament, the schedule will get re-generated.

Will close #168


features/step_definitions/content.rb
```diff
@@ -16,6 +16,13 @@
   expect(page).to have_link(text)
 end
 
+When(/^he (un)?checks the checkbox '(.*)'$/) do |un, checkbox_name|
```
DanielStephan97: What does un mean? Please use a descriptive name.

app/controllers/events_controller.rb
```diff
@@ -49,6 +49,7 @@ def create
 
   # PATCH/PUT /events/1
   def update
+    @event.invalidate_schedule if event_params[:has_place_3_match].to_i.zero? == @event.has_place_3_match
```
fpottbaecker, line 45: You sure this should happen here? Not in an `after_commit` hook on the model? You cannot be sure if the model will actually be saved, and you delete all matches. Even those with results.

db/migrate/20180117194839_add_place3_optional_flag.rb
```diff
@@ -0,0 +1,5 @@
+class AddPlace3OptionalFlag < ActiveRecord::Migration[5.1]
+  def change
+    add_column :events, :has_place_3_match, :boolean, default: true
```
fpottbaecker, line 3: Are you sure this is an acceptable name?

app/views/events/forms/_Tournament.html.erb
```diff
@@ -6,7 +6,13 @@
...
   <div class="col-lg-10">
     <%= form.select(:game_mode, Tournament.game_modes.map {|k, _| [Tournament.human_game_mode(k), k]},
-                 { selected: @event.game_mode, include_blank: t('events.new.select_game_mode')},
-                 { class: 'form-control' }) %>
+                    { selected: @event.game_mode, include_blank: t('events.new.select_game_mode')},
+                    { class: 'form-control' }) %>
+  </div>
+</div>
+<div class="form-group" id="event_has_place_3_match">
+  <div class="col-lg-10">
```
fpottbaecker: The label has to be outside the `col-lg-10`.

screen shot 2018-01-19 at 13 32 07

db/migrate/20180117201300_make_ko_default_tournament_mode.rb
```diff
@@ -0,0 +1,7 @@
+class MakeKoDefaultTournamentMode < ActiveRecord::Migration[5.1]
+  def up
+    change_column :events, :game_mode, :integer, default: 0, null: false
```
fpottbaecker: Um, this can be part of a change method.

fpottbaecker: And this influences all types of events, should you not put this elsewhere? Like, in the model?

features/step_definitions/content.rb
fpottbaecker: How about highlighting `(unchecked|checked)` and binding it to parameter `action`?

features/step_definitions/tournament.rb
```diff
@@ -41,10 +41,14 @@
   visit event_schedule_path(tournament_named tournamentName)
 end
 
-And(/^the texts? (.+) (?:are|is) there\.$/) do |texts_raw|
+And(/^the texts? (.+) (?:are|is)( not)? there\.?$/) do |texts_raw, _not|
```
fpottbaecker: why `_not`?

fpottbaecker: It might be better style and less code to have two actions, with a not or no not.

features/step_definitions/tournament.rb
```diff
@@ -111,3 +115,25 @@ def placing_to_display_string(placing)
...
+  for each in 1..num_teams do
+    single_tournament.add_team create_team
+  end
+end
+
+And(/^he fills in valid tournament data$/) do
+  fill_in :tournament_name, with: 'Dummy'
+  fill_in :tournament_discipline, with: 'DummySport'
+  fill_in :tournament_max_teams, with: 8
+  page.select I18n.t('activerecord.attributes.event.player_types.team'), from: :event_player_type
```
fpottbaecker: just use `select`

fpottbaecker: And you could look at the other selection tests in rspec for a cleaner way to select these. Maybe `:team` and `from: Event.human_attribute_name(...)`.

features/step_definitions/tournament.rb
```diff
@@ -111,3 +115,25 @@ def placing_to_display_string(placing)
...
+  fill_in :event_deadline, with: Date.today + 1.day
+  fill_in :event_startdate, with: Date.today + 2.day
+  fill_in :event_enddate, with: Date.today + 3.day
+  single_tournament
+  #fill_in "match_#{match.id}_match_points_away", with: points_away
```
fpottbaecker: Dead code

features/support/env.local.rb
```diff
@@ -106,6 +106,9 @@ def tournament_named(name)
   end
 
   def single_tournament
+    if @tournaments.empty?
```
fpottbaecker: Um, no. You should use a different method for form created objects.

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Tournament optional place 3 match
+  As a user
+  In order to easily create a tournament
```
fpottbaecker: I am not sure with the motivation here.

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
...
+  I want to make the place 3 match optional
+
+  Scenario: Create match without place 3 match
+    Given a user
+    And the user is logged in
+    When he visits the create event path
+    And he clicks 'New tournament'
```
fpottbaecker, line 10: You could compress this to "he creates a tournament"

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
...
+    And he fills in valid tournament data
+    And he unchecks the checkbox 'Spiel um Platz 3'
+    And he clicks 'Create'
+
+    Given 8 teams join the tournament
+    And the schedule page is visited
+
+    Then the text Spiel um Platz 3 is not there
```
fpottbaecker: Check the data instead of the page.

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
...
+    When he visits the create event path
+    And he clicks 'New tournament'
+    And he fills in valid tournament data
+    And he checks the checkbox 'Spiel um Platz 3'
+    And he clicks 'Create'
+
+    Given 8 teams join the tournament
+    And the schedule page is visited
+
+    Then the text Spiel um Platz 3 is there
```
Pottiman: Same here.

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
...
+  Scenario: Create match with place 3 match
+    Given a user
+    And the user is logged in
+    When he visits the create event path
+    And he clicks 'New tournament'
+    And he fills in valid tournament data
+    And he checks the checkbox 'Spiel um Platz 3'
+    And he clicks 'Create'
+
+    Given 8 teams join the tournament
```
fpottbaecker: This should be a "When"

app/views/events/forms/_Tournament.html.erb
```diff
@@ -6,7 +6,13 @@
   <%= form.label :game_mode, class: 'control-label col-lg-2' %>
   <div class="col-lg-10">
     <%= form.select(:game_mode, Tournament.game_modes.map {|k, _| [Tournament.human_game_mode(k), k]},
-                 { selected: @event.game_mode, include_blank: t('events.new.select_game_mode')},
-                 { class: 'form-control' }) %>
+                    { selected: @event.game_mode, include_blank: t('events.new.select_game_mode')},
```
fpottbaecker: Why did the indentation change?

fpottbaecker (changes requested):

app/controllers/events_controller.rb
ErikBrendel: Sounds good. How will this work?

db/migrate/20180117194839_add_place3_optional_flag.rb
ErikBrendel: it is acceptable. Maybe there are better names. Proposals?

db/migrate/20180117201300_make_ko_default_tournament_mode.rb
ErikBrendel: yes, but then the migration would not be revertable.

is it even possible to make it model side only?

features/step_definitions/tournament.rb
ErikBrendel: because `<space>not` and `not` are not valid variable names

db/migrate/20180117201300_make_ko_default_tournament_mode.rb
fpottbaecker: Pending discussion with PO

fpottbaecker: And even then, this migration is not reversible, you need to write a corresponding down method

features/step_definitions/tournament.rb
fpottbaecker: see comment below

features/support/env.local.rb
ErikBrendel: why?

features/tournament/optional_place_3_match.feature
ErikBrendel: I'm against that. That would hide the fact that here a button gets clicked

app/controllers/events_controller.rb
Askir: Checkout: https://apidock.com/rails/ActiveRecord/Transactions/ClassMethods/after_commit
Rails offers mutltiple hook methods so that you can perform actions on the model after specific actions (in this case update, create or destroy).
Goal should be to keep the controller somwhat lean ;)

app/controllers/events_controller.rb
```diff
@@ -227,6 +228,7 @@ def event_params
                                     :enddate,
                                     :initial_value,
                                     :gameday_duration,
+                                    :has_place_3_match,
```
Askir: I hate that this is getting so big but I have no idea how to fix it other than by creating multiple controllers for different event kinds. Which we agreed on not doing, because it would also mean that STI was probably not the right choice, which we didn't wanna change... arrrg.
Well, I suppose I can live with it.

db/migrate/20180117194839_add_place3_optional_flag.rb
Askir: `third_place_play-off` maybe? Other suggestions could be the equivalent of "kleines Finale" which I don't know if it exists in english :P

db/migrate/20180117201300_make_ko_default_tournament_mode.rb
Askir: Since you are just changing what is default I don't even think you need down and up. But yes in my opinion this should be part of the model, since maybe Rankinglists or Leagues don't need a default game_mode. You can implement this with another hook. This time `before_save`
Also you can validate the presence of a game_mode in the model which is another way of saying `null:false` although I think this was the case before you added this migration anyways.

Askir (commented):
In addition to the 4 comments:
Is it intentional that you guys only use cucumber for tests now? I thought it was supposed to be in addition to unit tests with rspec and just replaces feature tests. But you don't even test the changes in the controller action and also don't seem to test your own generate_schedule method at all.

app/controllers/events_controller.rb
Markus-Brand: Seems, like we don't have the information we need in a after_commit hook.

db/migrate/20180117194839_add_place3_optional_flag.rb
Markus-Brand: without the "has" it sounds like we are referencing a match object, not a boolean. And the difference between, place_3_match, third_place_match or third_place_play_off isn't too big, imo

features/tournament/optional_place_3_match.feature
fpottbaecker: How about "he visits the create tournament form"

features/tournament/optional_place_3_match.feature
```diff
@@ -0,0 +1,32 @@
...
+  I want to make the place 3 match optional
+
+  Scenario: Create match without place 3 match
+    Given a user
+    And the user is logged in
+    When he visits the create event path
+    And he clicks 'New tournament'
+    And he fills in valid tournament data
+    And he unchecks the checkbox 'Spiel um Platz 3'
+    And he clicks 'Create'
```
fpottbaecker: Could you use "he submits" instead. (as in: presses the button named `commit`)?

features/tournament/optional_place_3_match.feature
fpottbaecker: The data should be tested as well, which it is not, when this step fails, either the data is wrong or the view is incorrect, which is it?

db/migrate/20180124104518_add_default_start_time_to_matches.rb
```diff
@@ -1,5 +0,0 @@
-class AddDefaultStartTimeToMatches < ActiveRecord::Migration[5.1]
-  def change
-    change_column_default :matches, :start_time, -> { 'CURRENT_TIMESTAMP' }
```
fpottbaecker: You should not delete an existing migration that has been deployed. Please add a new migration changing the default again or [reverting this one](http://guides.rubyonrails.org/active_record_migrations.html#reverting-previous-migrations)

db/schema.rb
```diff
@@ -1,167 +0,0 @@
...
-#
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
-#
-# It's strongly recommended that you check this file into your version control system.
-
-ActiveRecord::Schema.define(version: 20180128145059) do
```
fpottbaecker: Why did you delete the schema? It has a purpose...

features/step_definitions/tournament.rb
```diff
@@ -115,3 +122,29 @@ def placing_to_display_string(placing)
...
+  fill_in :tournament_discipline, with: 'DummySport'
+  fill_in :tournament_max_teams, with: 8
+  select I18n.t('activerecord.attributes.event.player_types.team'), from: :event_player_type
+  select I18n.t('activerecord.attributes.tournament.game_modes.ko'), from: :tournament_game_mode
+  fill_in :event_deadline, with: Date.today + 1.day
+  fill_in :event_startdate, with: Date.today + 2.day
+  fill_in :event_enddate, with: Date.today + 3.day
+end
+
+And(/^he creates the tournament$/) do
```
fpottbaecker: When

features/step_definitions/tournament.rb
```diff
@@ -115,3 +122,29 @@ def placing_to_display_string(placing)
...
+  visit event_schedule_path(single_tournament)
+end
+
+Given(/^(\d+) teams join the tournament$/) do |num_teams|
+  for each in 1..num_teams do
+    single_tournament.add_team create_team
+  end
+end
+
+And(/^he fills in valid tournament data$/) do
```
fpottbaecker: When

features/step_definitions/tournament.rb
```diff
@@ -115,3 +122,29 @@ def placing_to_display_string(placing)
...
+  select I18n.t('activerecord.attributes.event.player_types.team'), from: :event_player_type
+  select I18n.t('activerecord.attributes.tournament.game_modes.ko'), from: :tournament_game_mode
+  fill_in :event_deadline, with: Date.today + 1.day
+  fill_in :event_startdate, with: Date.today + 2.day
+  fill_in :event_enddate, with: Date.today + 3.day
+end
+
+And(/^he creates the tournament$/) do
+  click_on I18n.t('helpers.submit.create', model: Tournament.model_name.human)
+  add_tournament Tournament.last
```
fpottbaecker: Optional: Check if the amount of tournaments increased and fail if not.

fpottbaecker (changes requested):

features/tournament/optional_place_3_match.feature
Markus-Brand: I don't think, that it would add any value. Can we just leave it as is?

fpottbaecker (approved):

Issue #356, opened 2018-01-16
Hide filtering of teams for signed off/unregistered users


svitlanakalancha:
Signed off/unregistered users should not be able to filter teams on the **Teams** page.

![sport-portal](https://user-images.githubusercontent.com/27727717/34983624-772ec142-faae-11e7-926b-0572dca1fc01.png)

lheilmann:
Needed time: 1h

PR #383, opened 2018-01-17
Hide filtering of teams for unsigned users

lheilmann:
See #356


db/schema.rb
```diff
@@ -44,6 +44,8 @@
   create_table "events_users", id: false, force: :cascade do |t|
     t.integer "event_id", null: false
     t.integer "user_id", null: false
+    t.index ["event_id", "user_id"], name: "index_events_users_on_event_id_and_user_id"
```
birneamstiel: I was wondering: Is there an explanation for why this is in here?

birneamstiel (approved):
LGTM!
(haven't tested yet)

db/schema.rb
lheilmann: No idea! Probably generated after `rails db:migrate`. What would you prefer? Ignore it or keep it out of this PR?

spec/views/teams/index.html.erb_spec.rb
```diff
@@ -39,4 +39,22 @@
     render
     expect(rendered).to have_selector(:link_or_button, t('teams.index.create_team'))
   end
+
+  it "renders the delete button when signed in and you are an admin" do
```
birneamstiel: @lheilmann we think this test doesn't belong here... It probably got in accidentally when resolving merge conflicts.

birneamstiel (changes requested):

spec/views/teams/index.html.erb_spec.rb
lheilmann: @birneamstiel You're right! Should I exlude it from this PR?

birneamstiel (approved):

PR #385, opened 2018-01-17
Kickerz/#337/enable admin to delete teams

lheilmann:


spec/models/ability_spec.rb
```diff
@@ -114,6 +114,51 @@
...
+  end
+
+  it 'should be able to destroy a team as an admin' do
+    ability = Ability.new(@admin)
+    team = FactoryBot.create :team
+
+    expect(ability).to be_able_to(:destroy, team)
+  end
+
+  it 'should not be able to destroy a team when participating in an event as a admin' do
```
birneamstiel: This should be '**should be** able to destroy a team when participating in an event as a admin', I think

app/controllers/teams_controller.rb
```diff
@@ -52,15 +52,11 @@ def update
...
-      team_id = @team.id
-      TeamUser.where(team_id: team_id).delete_all
-      @team.destroy
-      redirect_to teams_url, notice: I18n.t('helpers.flash.destroyed', resource_name: team_resource).capitalize
-    else
-      redirect_to @team, alert: I18n.t('helpers.teams.cant_destroy', resource_name: team_resource).capitalize
-    end
+    team_id = @team.id
+    TeamUser.where(team_id: team_id).delete_all
+    # Finally destroy team
```
birneamstiel: I would vote for removing this comment since `@team.destroy` is pretty descriptive, but it's your call :-)

birneamstiel (approved):
Looking good :-)

PR #389, opened 2018-01-18
Flexibus/#371/admin can edit all events

MariusDanner:
Please review @zhamanakov


spec/controllers/events_controller_spec.rb
```diff
@@ -142,6 +142,14 @@
       get :edit, params: { id: event.to_param }
       expect(response).to_not be_success
     end
+
+    it "should allow normal user to edit others created event" do
```
emanuel-metzenthin: You are not testing a normal user here by signing in an admin.

emanuel-metzenthin (changes requested):

emanuel-metzenthin (approved):

PR #392, opened 2018-01-18
325/bruteforce/event participants upper limit

### LINKED ISSUE #325: Upper limit participants in an event
Entering a too large number (like 1000000000000000000000000)in the "Anzahl Teams" Field will result in a crash of the creation process.

**Acceptance Criteria**

- [ ] The maximal upper limit of the Team count should be 1000 Teams for an Event.
- [ ] If a User enters a Number larger than 1000 they should be notified that the entered Number was to o large and needs to smaller than or equal to 1000.
- [ ] Should be ensured after editing an event as well
- [ ]  check whether leaving the field empty (player count "unbegrenzt") result in any problems, if yes make it an mandatory field

usame-algan:
closes #325


db/schema.rb
```diff
@@ -83,13 +83,6 @@
     t.index ["user_id"], name: "index_organizers_on_user_id"
   end
 
-  create_table "participants", force: :cascade do |t|
-    t.integer "attendee_id"
-    t.string "attendee_type"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
```
CoderWombat: I don't understand why this table is removed

CoderWombat: So i face the same problem in my Issue now maybe the Table just isn' t used anymore.

CoderWombat (approved):
Merge the dev into this and i think we can merge this

PR #393, opened 2018-01-18
279/bruteforce/group related fields

### LINKED ISSUE #279: group related fields in create league form
As creator of an event/league
I want related input fields to be grouped together
In order to help me understand the creation process better

**Group fields:**

Basisinformationen
- Name, Beschreibung, Sportart, Spielsystem

Teilnehmer
- Teilnahmeform, Anzahl Teams

Termine
- Dauer eines Spieltages
- Anmeldefrist, Beginn, Ende

**There may be changes after customer feedback.**

usame-algan:
closes #279


MarcelSchmidberger:
looks awesome but there are small changes necessary:

- the startvalue in the ranking list should be the last element in the list

image

db/schema.rb
```diff
@@ -83,13 +83,6 @@
     t.index ["user_id"], name: "index_organizers_on_user_id"
   end
 
-  create_table "participants", force: :cascade do |t|
-    t.integer "attendee_id"
-    t.string "attendee_type"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
```
Askir: This looks like you deleted a table? Once you merged in the dev see if this diff is gone. Otherwise check if migrating again fixes it.

app/assets/stylesheets/events.css
```diff
@@ -3,7 +3,47 @@
...
+#event_gameday_duration_div {
+    order:4;
+}
+
+#event_deadline_div,
+#event_startdate_div,
+#event_enddate_div,
+#event_duration_div,
+#event_submit_div {
+    order:5;
```
Askir: Looks quite cool to me although having an ID for every single form element seems a little rigid.

Askir (changes requested):
Looks quite fine to me, just merge in dev and I think Erik has a few comments on the css.

app/assets/stylesheets/events.css
erksch: I would suggest creating reusable classes for the order property.
When using SCSS you could write something like
```
@for $i from 1 through 5 {
    .order-#{$i} { order: $i; }
}
``` 
and assign the classes to the form-groups.
It prevents creating many many ids.
If you have a different opinion what to use here, i am looking forward to discuss it with you 😄

erksch (changes requested):

app/assets/stylesheets/events.css
erksch: Of course, with this approach you need to change the file name from `.css` to `.scss` (it should work then hopefully).

usame-algan: This is a really good idea and should work! Didn't know we could use scss in our project 😆

Askir (approved):
I like it a lot :)

MarcelSchmidberger:
I know this isnt in PO Review yet but I realized I mad a mistake in my requirements. It would be awesome if you could change the position of "Spieltagdauer" to be behind "Ende". Sorry for this mistake of mine. 

image

Next the "Erweitert button" appears as first element, but should actually be the last one of the entire form. So it should be just ahead of the create or cancel buttons. 

image

erksch (approved):

PR #394, opened 2018-01-18
Bruteforce/#324/add sporttype dropdown in event creation

CoderWombat:


Askir (approved):
looks good to me :) Tested functionality and it seems to be working as expected. Any info on cucumber tests for this? Also you could add a unit test that checks if there is a dropdown for the view.

app/views/events/_form.html.erb
```diff
@@ -41,7 +41,15 @@
   <div class="form-group">
     <%= f.label :discipline, class: 'control-label col-lg-2' %>
     <div class="col-lg-10">
-      <%= f.text_field :discipline, class: 'form-control' %>
+      <%= f.text_field :discipline, list: "discipline", class: 'form-control' %>
+      <datalist id = "discipline">
```
usame-algan, line 45: Interesting choice! I quickly checked the browser support for this element and unfortunately it is not supported in Safari and iOS Safari and only partially supported in Firefox and Edge. See: https://caniuse.com/#feat=datalist
We could use jQuery UIs autocomplete feature instead: http://jqueryui.com/autocomplete/

PR #398, opened 2018-01-18
Fix menu highlighting. Remove redundant menu entry

ErikBrendel:
With this commit, the weird highlighting gets fixed.

It occured, since `users/:id` is a sub-path from `users/:id/dashboard` (for example), and bootstrap thought that we are currently on both pages (which is technically not wrong :D)

I fixed this by **changing the** `users#show` **- URL** from `/users/:id` to `/users/:id/profile`.

-------

In addition to that, the **Menu - Entry "Mein Account" got removed**, since it does exactly the same as "Profil bearbeiten"


GittiHab:
Why am I added to review this? I am not a developer.

ErikBrendel:
@GittiHab I just wanted to let you know that this issue got addressed. Of course you do not have to review code :D

config/routes.rb
```diff
@@ -36,6 +36,10 @@
   # Sets the devise scope to be used in the controller.
   # http://www.rubydoc.info/github/plataformatec/devise/ActionDispatch%2FRouting%2FMapper%3Adevise_scope
   devise_scope :user do
+    get '/users/:id/profile', to: 'users#show', as: 'user'
+    get '/users/:id', to: redirect('/users/%{id}/profile')
```
fpottbaecker, line 50: all of these routes should be member routes, to avoid the necessity of `id`s here

fpottbaecker (changes requested):

config/routes.rb
ErikBrendel: What does this mean? Would you be so kind to adjust it directly on this branch?

Markus-Brand (approved):

fpottbaecker (approved):

config/routes.rb
```diff
@@ -41,14 +41,18 @@
   # Sets the devise scope to be used in the controller.
   # http://www.rubydoc.info/github/plataformatec/devise/ActionDispatch%2FRouting%2FMapper%3Adevise_scope
   devise_scope :user do
+    get '/users/:id', to: redirect('/users/%{id}/profile')
```
fpottbaecker: One note though: Is this redirect necessary, if all links already link to `/users/:id/profile`?

ErikBrendel: It should not be needed directly. It is convenient, tough (for devs that got used to the url and type it in manually and to keep possible references from the outside intact)

PR #401, opened 2018-01-19
Flexibus/#379/teams view as tiles

### LINKED ISSUE #379: Teams view as tiles
**#380 depends on this for tile component**
**#374 depends on this for tile component and team logic**

New Layout for "Teams" Page:

If the team has a team-picture, then tile has this picture. If not, then we use 1 of 2 default pictures (they can be easily found by link in Flexibus channel)

If width for content is <=800px, then it's only 2 blocks in a row. (**this is not shown in pictures**)

- [ ] create a reusable component for the tiles that can be used in #380.

- whole tile should be clickable
- no priority for teams
- 'Deine Teams' only shows teams that user is a member of
- 'Deine Teams' is subset of 'Alle Teams' (not disjunct)

- 'Deine Teams' block is not shown when there is no data

emanuel-metzenthin:
Closes #379.


app/assets/stylesheets/tile.css
```diff
@@ -0,0 +1,49 @@
+@media screen and (min-width : 800px) {
+    .tile-wrapper {
+        display: grid;
```
usame-algan, line 3: How do we handle browser support? See https://caniuse.com/#feat=css-grid
Do we have a fallback to flexbox or something similar?

app/views/application/_tile_template.html.erb
```diff
@@ -0,0 +1,11 @@
+<div onclick="location.href='<%= destination_path%>';" class="tile">
```
usame-algan: This won't work if Javascript is disabled in the Browser. Could use an `<a>` Tag with absolute positioning to achieve the same result.

app/assets/stylesheets/tile.css
emanuel-metzenthin: Grid is not compatible with Opera Mini and UC Browser for Android. All major browsers support it. So I guess we're fine :).

usame-algan (changes requested):

app/views/application/_tile_template.html.erb
emanuel-metzenthin: I think many features are dependent of JS already ;). Nonetheless I changed it,  good point. Could you accept?

usame-algan (approved):

PR #405, opened 2018-01-19
Bruteforce/#277/event tests refactoring

Askir:
Hoping that this will make the tests a little bit less rigid.

### DIFF
--- a/spec/factories/leagues.rb
+++ b/spec/factories/leagues.rb
@@ -2,10 +2,5 @@ FactoryBot.define do
   factory :league, parent: :event, traits: [:has_dates], class: League do
     game_mode :round_robin
     gameday_duration 7
-
-    factory :league_with_teams, class: League, parent: :event_with_teams, traits: [:has_dates] do
-      game_mode :round_robin
-      gameday_duration 7
-    end
   end
 end
--- a/spec/features/event/new_event_spec.rb
+++ b/spec/features/event/new_event_spec.rb
@@ -2,15 +2,28 @@ require 'rails_helper'
 
 describe 'new event page', type: :feature do
 
-  def fill_in_creation_form_with(fields = {})
+  def fill_in_fields_with(fields = {})
     fields.each do |field_symbol, input|
       fill_in Event.human_attribute_name(field_symbol), with: input
     end
   end
 
+  def select_dropdowns_with(dropdowns = {})
+    dropdowns.each do |dropdown, input|
+      select input, from: dropdown
+    end
+  end
+
   def has_input_fields(*fields)
     fields.each do |field|
-        expect(page).to have_field(Event.human_attribute_name field)
+      expect(page).to have_field(Event.human_attribute_name field)
+    end
+  end
+
+  def has_content(fields)
+    fields.each do |field, content|
+      expect(page).to have_content(Event.human_attribute_name field)
+      expect(page).to have_content content
     end
   end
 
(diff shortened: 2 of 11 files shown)


spec/factories/leagues.rb
```diff
@@ -2,10 +2,5 @@
   factory :league, parent: :event, traits: [:has_dates], class: League do
     game_mode :round_robin
     gameday_duration 7
-
-    factory :league_with_teams, class: League, parent: :event_with_teams, traits: [:has_dates] do
```
erksch, line 6: why are you removing this?

spec/features/event/new_event_spec.rb
```diff
@@ -10,90 +36,89 @@
...
+      visit new_path
+    end
 
     it 'should be possible to enter date conditions' do
       visit new_path
 
-      expect(page).to have_field(Event.human_attribute_name :deadline)
-      expect(page).to have_field(Event.human_attribute_name :startdate)
-      expect(page).to have_field(Event.human_attribute_name :enddate)
+      has_input_fields :deadline, :startdate, :enddate
```
erksch, line 45: Sweet!

spec/factories/leagues.rb
Askir: Because you can just use the new trait :with_teams instead and combine features of event objects as you wish

spec/models/event_spec.rb
```diff
@@ -23,62 +23,21 @@
...
 
   it 'should not validate without discipline' do
-    league = FactoryBot.build(:league, discipline: nil)
-    expect(league.valid?).to eq(false)
+    event = FactoryBot.build(:event, discipline: nil)
+    expect(event.valid?).to eq(false)
   end
 
   it 'should not validate without game_mode' do
-    league = FactoryBot.build(:league, game_mode: nil)
```
erksch, line 39: Why are you removing all these tests?

Askir: event_spec shouldn't test things that are league and tournament specifics. So I extracted them into shared_examples see further down in the files list.

spec/factories/leagues.rb
erksch: Ah I got it. Pretty sweet with new trait.

spec/models/event_spec.rb
Askir: Actually see further up in actual_events_example.rb ;)

erksch: ok and where are the league specific tests that are not included in the shared_example?

Askir: can you clarify? I'm pretty sure I got all of them in the shared example

erksch: My question is if all the tests where general event tests or if there are any specific league or tournament tests left which you did not move into the shared example?

spec/models/event_spec.rb
```diff
@@ -23,62 +23,21 @@
...
 describe 'Event model', type: :model do
 
-  let(:event) { FactoryBot.build(:league) }
+  let(:event) { FactoryBot.build(:event) }
 
   it 'should not validate without name' do
-    league = FactoryBot.build(:league, name: nil)
-    expect(league.valid?).to eq(false)
+    event = FactoryBot.build(:event, name: nil)
+    expect(event.valid?).to eq(false)
```
erksch, line 30: Why is this test not in the shared example?

spec/models/event_spec.rb
Askir: Well there are specific league and tournament tests but they are in league_spec.rb and tournament_spec.rb like they are supposed to.

spec/models/event_spec.rb
Askir: because it is validated in the events model and not in the league or tournament model.
`validates :name, :discipline, :game_mode, presence: true ` is the first line in the Event class. All these validations are tested in Event_spec.rb

erksch (approved):

Issue #294, opened 2018-01-07
Display Events as Tiles


PoJo93:
AS a User
I WANT TO see on the event page events in a visual layout of tiles
TO get a more visual appealing layout

ACCEPTANCE CRITERIA
- [ ] On the event page I see all events as tiles
   - Every tile has a picture 
   - Every tile has the event name
   - Every tile has a date. It is the next date which is approaching (Signup, Begin, End). Depending on the date it should also have the corresponding label
   - Every tile has the kind of Sport
   - Every tile has how many teams/participants are already signed up and also how many could be there in total
- [ ] In the settings of the event I have a button to upload, change and delete a picture (Like on the profile page)
- [ ] When I create an event I want to have the option to upload, change and delete a picture (Like on the profile page)
- [ ] Every event has a default picture (source will be linked here soon)
- [ ] When I click on the tile it should direct me to the corresponding event

Similar to:

![screen shot 2018-01-07 at 21 26 41](https://user-images.githubusercontent.com/5064619/34653797-9c0248a4-f3f1-11e7-80a9-edc204fbb714.png)

MarcelSchmidberger:
This may look like this 
![image](https://user-images.githubusercontent.com/29160369/34655615-c017875c-f40c-11e7-9c8f-abff416a7d27.png)

Please feel free to change the Information amount on the tile (remove or add fields)

nikriek:
Prio: Picture upload.

nikriek:
@PoJo93 Just to check with you again: I should only add image data field to events + the upload actions, right?

nikriek:
@PoJo93 The default image is still missing :)

PoJo93:
Yes just put the options for picture upload and the default image. The rest should be handled by the team of @zhamanakov :)

nikriek:
Real complexity: 8

nikriek:
Real complexity: 8

MariusDanner:
Is this ticket finished? Because we need the logic for #380

jkbe:
@nikriek

nikriek:
The upload and an image assigned to an event is in place. No one reviewed it so far... you could do it :)

nikriek:
It’s merged already. What else do you need?

PoJo93:
Nice good work! The only thing missing is the default image. Could you include that image as default
[...]

PR #409, opened 2018-01-20
Add picture upload for events

nikriek:
Connects to #294


manner (approved):
Default image is still missing, but should be handled by other team

Issue #139, opened 2017-11-26
Swiss system support


MarcelSchmidberger:
As a event organizer
I want to create an event using the swiss system for match making
In order to ensure evenly matched player/team strengths in the games

https://de.wikipedia.org/wiki/Turnierform
https://de.wikipedia.org/wiki/Schweizer_System

Depends on implementation of #125 

**Acceptance criteria**
- [x] Selecting  "Schweizersystem" in create event Page #19 will create an event using the Swiss system
- [x]  After each game-day ("Spieltag") the upcoming matches for the next game-day will automatically be calculated
- [ ] These matches are calculated according to the swiss system
   - [x] Matches for the first game-day are random
   - [ ] For the next game-day the leading player plays against the second best, and so on
   - [ ] if there is no clear choice for a contender(for example multiple players ranked second), the selection is random
    - [ ] for starters there should be as many game-days as in a "jeder gegen jeden" so n-1 for n players
    - [ ] No two players play against each other twice

MarcelSchmidberger:
whats the criteria for calculating the new gameday right know? I assume the new gameday is calculated as soon as the new gameday starts (so 00:01 Uhr of the startdate of the gameday)? This results in problems when I want all gamedays to happen at the same time.

![image](https://user-images.githubusercontent.com/29160369/35816348-c00244ba-0a9a-11e8-8b70-4020bbe779d6.png)

Now all gamedays are calculated at once, so the order of the "Spielpaarungen" is random. 
Would it be possible to add a button "calculate new gameday" which takes the ranking as it is at this moment and based on this ranking the new gameday is calculated? This button should only be visible by the admin or organizer. Th button is an additional means to calculate the new gameday. The new gamedays are calculated when the new gameday starts or the next gameday can be generated by pressing the button. How much work would this be to implement?

This could look like this

![image](https://user-images.githubusercontent.com/29160369/35816675-9b419bd4-0a9b-11e8-96a1-6aa132f80514.png)

MarcelSchmidberger:
there seems to be a bug, when reloading the playing schedule after first generation results in an empty gameday 7 and all matches from this gameday are moved to gameday 8

![image](https://user-images.githubusercontent.com/29160369/35816850-11f9c882-0a9c-11e8-8e37-c9d31d40686a.png)

MarcelSchmidberger:
[...]

PR #410, opened 2018-01-20
bruteforce/#139/Swiss system support

erksch:
connect to #139


app/models/event.rb
```diff
@@ -67,6 +67,10 @@ def send_mails_when_canceled
     end
   end
 
+  def is_up_to_date
+    true
+  end
```
Askir, line 74: Why is this neseccary? Shouldn't the method only be called for leagues?

app/models/event/ranking.rb
```diff
@@ -22,7 +22,7 @@ def get_ranking
       #   1. DESCENDING by points
       #   2. DESCENDING by goal difference
       #   3. ASCENDING by name
-      @ranking_entries = @ranking_entries.sort_by { | ranking_entry | [-ranking_entry.points, -ranking_entry.goals_difference, ranking_entry.name] }
+      @ranking_entries = @ranking_entries.sort_by { | ranking_entry | [-ranking_entry.points, -ranking_entry.goals, ranking_entry.team.name] }
```
Askir: have you talked with POs about this? Why should the ranking be sorted by goals instead of gaol_difference?

app/models/event/ranking_entry.rb
```diff
@@ -2,7 +2,7 @@ class Event
   # Smallest unit of a ranking. Contains different data for one team and contains logic to parse this data out of
   # match objects.
   class RankingEntry
-    attr_reader :name, :match_count, :won_count, :draw_count, :lost_count, :goals, :goals_against, :goals_difference, :points
+    attr_reader :team, :match_count, :won_count, :draw_count, :lost_count, :goals, :goals_against, :goals_difference, :points
```
Askir, line 5: maybe just implement a getter yourself with:
``` Ruby
def name
  @team.name
end
```
but fine either way

app/models/league.rb
```diff
@@ -42,6 +42,70 @@ def generate_schedule
       calculate_round_robin
     elsif game_mode == League.game_modes.key(1)
       calculate_two_halfs
+    elsif game_mode == League.game_modes.key(2)
```
Askir: I liked it more when we were actually checking for the symbols here, any reason we changed that?
` game_mode == :swiss_system ` seems way more telling imo.

app/models/league.rb
```diff
@@ -42,6 +42,70 @@ def generate_schedule
...
+      calculate_swiss_system_start
+    end
+  end
+
+  def is_up_to_date
+    if game_mode != League.game_modes.key(2)
+      return true
+    end
```
Askir:
```Ruby
true unless game_mode == :swiss_system
```
is more rubylike

app/models/league.rb
```diff
@@ -42,6 +42,70 @@ def generate_schedule
...
+
+    gamedays.last.endtime > DateTime.now
```
Askir: try to use current instead of now, I think it's what we use everywhere else.

app/models/league.rb
```diff
@@ -42,6 +42,70 @@ def generate_schedule
...
+      end
+
+      if team_away_index < teams_amount
+        add_match(team_home, team_away, gameday_number)
+        teams_matched << team_home
+        teams_matched << team_away
+      end
+    end
+    save
+  end
```
Askir: I'm quite sure that the way you implemented the system it is possible to lock with a suboptimal matchday, where 2 teams that have already played against each other are left out and that way don't have a game at that gameday.

In [this stackoverflowpost](https://stackoverflow.com/questions/28629235/swiss-tournament-pairing-algorithm) you can find and example of a ranking which would end badly with this implementation.

See if you find any possilbe solution in the comments otherwise honestly just bruteforce it participant counts shouldn't be too high to actually matter for this I think.

app/models/league.rb
```diff
@@ -42,6 +42,70 @@ def generate_schedule
...
+
+  # create a random first gameday for the swiss system
+  def calculate_swiss_system_start
+    add_gameday
+    temp = teams.to_a.shuffle
+    temp.each do |team|
+      team_away = temp[1 + rand(temp.size - 2)]
+      add_match(team, team_away, 0)
+      temp.delete(team)
+      temp.delete(team_away)
```
Askir: you are editing the array while you are iterating over it, are you sure that this doesn't result in trouble?

app/views/matches/show.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <dl class="dl-horizontal">
   <dt><strong><%= model_class.human_attribute_name(:gameday) %>:</strong></dt>
-  <dd><%= @match.gameday %></dd>
+  <dd><%= @match.gameday_number %></dd>
```
Askir, line 8: Thank you, I would've hat to fix this otherwise :innocent:

spec/controllers/events_controller_spec.rb
```diff
@@ -358,18 +358,30 @@
...
+      get :schedule, params: { id: league.to_param }, session: valid_session
+      expect(league.matches).not_to be_empty
+    end
+
+    it "should update schedule if it is not up-to-date" do
+      league.gamedays << FactoryBot.create(:gameday)
+
+      last_gameday = league.gamedays.last
+      last_gameday.starttime = Date.today.next_day
+      last_gameday.endtime = Date.today.next_day 2
```
Askir: Again I'd prefer to use `Date.current +1.day` or `Date.current + 2.days` here.

spec/controllers/leagues_controller_spec.rb
```diff
@@ -74,7 +74,7 @@
           @league.add_team(team2)
           get :ranking, params: { id: @league.to_param }, session: valid_session
           ranking_entries = controller.instance_variable_get(:@ranking_entries)
-          expect(ranking_entries.first.name).to be < ranking_entries.second.name
+          expect(ranking_entries.first.team.name).to be < ranking_entries.second.team.name
```
Askir, line 77: for how often you had to edit that, a getter for the name doesn't seem like such a bad idea ;)

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
     end
+
+    context 'swiss system' do
+      let(:gamemode) { League.game_modes[:swiss] }
+
+      it 'does not have an up-to-date schedule if the last gameday is in the past' do
+        league.gamedays << FactoryBot.create(:gameday)
+
+        last_gameday = league.gamedays.last
+        last_gameday.starttime = 2.days.ago
```
Askir: another way of specifying dates? :D Yeh, maybe we should have a talk about this :P

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+
+        expect(league.have_already_played(teams[0], teams[1])).to be true
+        expect(league.have_already_played(teams[1], teams[0])).to be true
+      end
+
+      context 'update schedule' do
+        it 'creates a new gameday' do
+          gameday_amount = league.gamedays.length
+          league.update_schedule
+          expect(league.gamedays.length).to be gameday_amount + 1
```
Askir: have a look at the `change` matcher from rspec makes this a little bit more easily readable

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+        end
+
+
+        it 'creates a correct amount of matches' do
+          matches_amount = league.matches.length
+          league.update_schedule
+          expect(league.matches.length).to be matches_amount + (league.teams.length / 2).floor
```
Askir: This test only checks for one gameday with one constellation of Teams. The problem I mentioned earlier isn't tested this way.

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+          league.update_schedule
+          current_pairings = league.matches.map { |match| Set[match.team_home, match.team_away] }
+          expect(current_pairings.uniq.length == current_pairings.length).to be true
+        end
+
+        it 'creates the matches correctly by the ranking' do
+          league = FactoryBot.create(:league)
+          league.game_mode = League.game_modes[:swiss]
+          teams = FactoryBot.create_list(:team, 6)
+          league.teams.append teams
```
Askir: you can use `<<` instead of `append` I think

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+          league.gamedays << FactoryBot.create(:gameday)
+
+          match1 = FactoryBot.create(:match, team_home: teams[0], team_away: teams[1], gameday_number: 1,
+                                      points_home: 3, points_away: 0)
```
Askir: Do we need to specify points here? The test really does look quite ugly, try to reduce the amount of variables needed for the matches as much as possible.

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+          match1.game_results << FactoryBot.build(:game_result,
+                                                  score_home: 20, # team 0
+                                                  score_away: 0   # team 1
+          )
+          league.matches << match1
```
Askir: If you are using gamedays then there is no need to add the match manually to `league.matches`. Just add it to the matches of the gameday and refer to matches in your code via `league.all_matches`.

spec/models/league_spec.rb
```diff
@@ -140,6 +141,146 @@
...
+          )
+          league.matches << match2
+
+          match3 = FactoryBot.create(:match, team_home: teams[4], team_away: teams[5], gameday_number: 1,
+                                     points_home: 3, points_away: 0)
+          match3.game_results << FactoryBot.build(:game_result,
+                                                  score_home: 15, # team 4
+                                                  score_away: 3   # team 5
+          )
+          league.matches << match3
```
Askir: at least extract all this code into an extra method inside the test file so we have a method name that says what it does please :) and maybe put this in a let statement because it has nothing to do with what you are testing at all.

Askir (changes requested):
Quite a few minor things and some that are a little bit bigger :) This is definitely not done yet, sorry.

Tell me if you need help (:

app/models/event/ranking.rb
erksch: I think this is a just a mistake that occured while rebasing :) Thanks for finding it.

app/models/event.rb
erksch: Yes, but the benefit is clearer controller action. 
```
if @event.matches.empty?
       @event.generate_schedule
       @event.save
elsif not @event.is_up_to_date
      @event.update_schedule
      @event.save
end
```
I eliminate conditional statements: checking if the event is a league and the game_mode is the swiss system.

app/models/event/ranking_entry.rb
erksch: Does not seem really neccessary.
There were just a few calls of `ranking_entry.name`.
So I decided to replace all occurences with `ranking_entry.team.name` without much effort.

app/models/league.rb
erksch: It was not me, I just adopted the current design for consistency.
But you are right this should be changed.

app/models/league.rb
erksch: But wouldn't this return `true` or `false` right away? 
I want to exclude all other game_modes and then check for the condition of the swiss system.

app/models/league.rb
erksch: Okay.

app/views/matches/show.html.erb
erksch: You're welcome :)

app/models/league.rb
erksch: It worked, but to prevent further discussion, I think I will change this to an approach without deleting from the iterated array.

spec/controllers/events_controller_spec.rb
erksch: Added to the list.

spec/controllers/leagues_controller_spec.rb
erksch: Not that much ;)
And now the work is done.

spec/models/league_spec.rb
erksch: I think this is the most readable syntax, why not?

spec/models/league_spec.rb
erksch: You're right.

spec/models/league_spec.rb
erksch: I will try it without the points.

spec/models/league_spec.rb
erksch: Where is the difference between calling `league.matches << match1` and `gameday.matches << match1`?

spec/models/league_spec.rb
erksch: Does not seem much prettier :D
What do you want to extract? Creating all the matches?

app/models/league.rb
erksch: Do you mean `(return true) unless ...`?

app/models/event/ranking.rb
MarcelSchmidberger: @Askir yes your are right, the ranking should be sorted by the goal difference

app/models/league.rb
usame-algan: We changed the way we store `game_mode` from strings to integer a while ago so this was the solution. I can see that this is harder to read and should be changed. Maybe @Pottiman can help with this.

spec/models/league_spec.rb
erksch: This is just sweet. 
`expect{ league.update_schedule }.to change { league.gamedays.length }.by 1`

spec/controllers/events_controller_spec.rb
usame-algan: Is there a difference to `Date.tomorrow`?

app/models/league.rb
erksch: What about `League.game_modes[:swiss]`

app/models/league.rb
```diff
@@ -109,6 +127,43 @@ def round_robin_pairings(teams_array)
...
+          teams_matched << team_home
+          teams_matched << team_away
+        end
+      end
+    end
+
+    # create a random first gameday for the swiss system
+    def calculate_swiss_system_start
+      add_gameday
+      shuffledTeams = teams.to_a.shuffle
```
erksch: whups sorry for the camel case here.

app/models/event.rb
Askir: alright fine with me

spec/models/league_spec.rb
Askir: I just never know which versions are timezone dependent and which aren't so if we mix them up, they'll lead to issues at some point.

spec/models/league_spec.rb
Askir: ideally at some point we can remove the matches association from events when every feature uses gamedays instead. So I'd prefer it if the tests also do so, because that we can actually clean those relations up.

spec/models/league_spec.rb
Askir: yes, inside the it statement should actually only be the code that you are testing. You aren't really testing how match creation works or if you can add it to the matches of leagues, so it should be in a before block or rather a lazy evaluated let.

Also extracting this into a method enabled you to use it later or use it again in a test with a different ranking if you parametrize the method smarty.

app/models/league.rb
```diff
@@ -38,13 +38,31 @@ class League < Event
   enum game_mode: [:round_robin, :two_halfs, :swiss, :danish]
 
   def generate_schedule
-    if game_mode == League.game_modes.key(0)
+    if game_mode == League.game_modes[:round_robin]
       calculate_round_robin
-    elsif game_mode == League.game_modes.key(1)
+    elsif game_mode == League.game_modes[:two_halfs]
       calculate_two_halfs
+    elsif game_mode == League.game_modes[:swiss]
```
Askir: you should be able to access enum state via autogenerated accessors. `self.swiss?` should return true if the gamemode of this league is swiss. Try it out, more info here: https://hackhands.com/ruby-on-enums-queries-and-rails-4-1/

spec/models/league_spec.rb
erksch: Ah yeah alright.

app/models/league.rb
erksch: Wow nice :) I'll do that.

MarcelSchmidberger:
there seems to be a problem with the ranking table in your branch. I created an swiss system league and when visiting the ranking following error occurs:

image

Askir:
Good find @MarcelSchmidberger interesting that all tests run...

MarcelSchmidberger:
The playing schedule of the event mentioned above locked like this
image

Note here that matches between the same people occur multiple times.  
As this article describes the swiss system ass follows: https://de.wikipedia.org/wiki/Schweizer_System "Es wird ausgeschlossen, dass zwei Spieler zweimal aufeinandertreffen." 
To be a "true" swiss system it would be nice if new gamedays would not be generated if everybody played against everybody else already. But on the other hand if the organizer is able to achieve the event he can end it, when ever he wishes to. I check whether thats a bug or a feature (multi round swiss system would a good marketing name)

MarcelSchmidberger:
@Askir This isnt a good example for the quality of the tests

Askir:
@MarcelSchmidberger try with a few more members. It doesn't really make sense to let people keep playing after they have all played. The system wont work in that case for now. And for the first two gamedays it looks correct

MarcelSchmidberger:
yes the first gamedays are correct. And as far as I can see The playing schedule is correct until the point where everybody has played with everybody else. So the swiss system works correctly. It only needs to stop producing game days when everybody has played against everybody. But I will test with more players

app/models/event/ranking_entry.rb
Askir: As @MarcelSchmidberger mentioned you forgot one line ;) Which breaks the rankinglist. #plzfix

app/models/league.rb
```diff
@@ -109,6 +127,42 @@ def round_robin_pairings(teams_array)
       games.map { |game| game.select { |match| !match[1].nil? } }
     end
 
+    def penalty(ranked_teams, matches)
+      (return Float::INFINITY) if matches.flatten.uniq.length != matches.flatten.length
```
Askir: pretty sure these braces aren't neseccary

app/models/league.rb
```diff
@@ -109,6 +127,42 @@ def round_robin_pairings(teams_array)
...
+      matches.inject(0) { |sum, match| sum + penalty_for_match(ranked_teams, match) }
+    end
+
+    def penalty_for_match(ranked_teams, match)
+      (return Float::INFINITY) if have_already_played(match[0], match[1])
```
Askir: also here probably

app/models/league.rb
```diff
@@ -109,6 +127,42 @@ def round_robin_pairings(teams_array)
...
+    def calculate_swiss_system_new_gameday
+      add_gameday
+      gameday_number = gamedays.length - 1
+      ranking = get_ranking
+      ranked_teams = ranking.map(&:team)
+
+      all_matches = ranked_teams.to_a.combination(2).to_a.combination((ranked_teams.length / 2).floor)
+      best_matches = all_matches.reduce do |best_matches, current_matches|
+        (penalty(ranked_teams, best_matches) < penalty(ranked_teams, current_matches)) ? best_matches : current_matches
+      end
```
Askir: this could also probably be done with a min_by which might be a bit more readable than reduce.

spec/controllers/events_controller_spec.rb
```diff
@@ -358,18 +358,31 @@
...
-      event.add_participant(@user)
-      event.add_participant(@other_user)
-      event.generate_schedule
-      get :schedule, params: { id: event.to_param }, session: valid_session
-      expect(event.matches).not_to be_empty
+      league.add_participant(@user)
+      league.add_participant(@other_user)
+      league.generate_schedule
+      get :schedule, params: { id: league.to_param }, session: valid_session
+      expect(league.matches).not_to be_empty
```
Askir: use all_matches plz :D

spec/models/league_spec.rb
```diff
@@ -140,6 +141,163 @@
...
+            match1.game_results << FactoryBot.build(:game_result,
+                                                    score_home: 25, # team 0
+                                                    score_away: 5  # team 1
+            )
+            match2 = FactoryBot.create(:match, team_home: teams[2], team_away: teams[3], gameday_number: 1, points_home: 3, points_away: 0)
+            match2.game_results << FactoryBot.build(:game_result,
+                                                    score_home: 26, # team 2
+                                                    score_away: 2   # team 3
+            )
+            match3 = FactoryBot.create(:match, team_home: teams[4], team_away: teams[5], gameday_number: 1, points_home: 3, points_away: 0)
```
Askir: is crate neseccary here?

spec/models/league_spec.rb
```diff
@@ -140,6 +141,163 @@
...
+            league.matches << gameday.matches
+            league.update_schedule
+          end
+
+          let(:sum_distances) {
+            ranked_teams = league.get_ranking.map(&:team)
+            league.gamedays.last.matches.inject(0) { |sum, match|
+              sum + (ranked_teams.index(match.team_home) - ranked_teams.index(match.team_away)).abs
+            }
+          }
```
Askir, line 248: duplication between test code and production code. I really don't like this. Imo just make the method public.

Askir (changes requested):
very small issues but blocking so that you don't forget to fix the rankinglist page :) Also please add a show test for it while you're at it. So that nobody else breaks it by accident in the future. Just check if it renders. Everything else is nice but not neseccary for now.

Askir (approved):

PR #411, opened 2018-01-21
Faster OpenID Signup

### LINKED ISSUE #373: faster OpenID signup
**As an** individual signing up with OpenID
**I** do not **want to** have to enter my Name during the registration
**In order to** have no unnecessary effort.

##Acceptance Criteria:
  * [x] if you sign up with OpenID you shouldn't have to enter additional data since the platform should be able to receive the necessary data

workflow:
openidlogin

fpottbaecker:
Closes #373


fpottbaecker:
Reviewed and accepted by PO

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -30,11 +30,13 @@ def sign_in_user(user)
       flash_if_navigational :notice, :success, provider: 'OpenID'
     end
 
-    def sign_up_user(auth)
+    def sign_up_user(auth, user)
```
ErikBrendel: why the additional parameter? It does not look used to me

app/models/user.rb
```diff
@@ -122,6 +124,9 @@ def valid_omniauth_session?(session)
     def from_omniauth(auth)
       where(provider: auth.provider, uid: auth.uid).first_or_create do |user|
         user.email = auth.info.email
+        user.first_name = auth.info.first_name
+        user.last_name = auth.info.last_name
+        user.password = Devise.friendly_token 32
```
ErikBrendel: why 32? Could this be extracted to a constant?

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -60,13 +60,25 @@
     end
 
     context 'given no (logged in or linked) user' do
-      it 'should redirect to the registration from' do
+      it 'should redirect to the registration from if information is missing' do
```
ErikBrendel: form?

ErikBrendel (changes requested):
Where exactly happens the logic `if names_given redirect_to_dashboard else redirect_to_signup_form` ?

Oluwoye (approved):

app/controllers/users_controller.rb
```diff
@@ -120,7 +120,7 @@ def admin_update_params
     end
 
     def generate_random_password
-      token = Devise.friendly_token 32
+      token = Devise.friendly_token User::OMNIAUTH_PASSWORD_LENGTH
```
ErikBrendel, line 123: what does friendly mean here? Are there also cryptic tokens?

ErikBrendel (approved):

app/controllers/users_controller.rb
fpottbaecker: As far as i know, that generates a base64 string, which can be saved as ASCII, but is random.

PR #412, opened 2018-01-21
added seperate match  edit view for admin and other users

akuchinke:


app/views/matches/_form.html.erb
```diff
@@ -13,21 +13,34 @@
   <div class="form-group">
     <%= f.label :team_home_id, for: "team_home_id", :class => 'control-label col-lg-2' %>
     <div class="col-lg-10">
+      <% if current_user!=nil and current_user.admin==true %>
```
nikriek: Please use `nil?` or `blank?` and maybe add a helper function instead of the ugly `== true`

nikriek (changes requested):

citron0xa9 (approved):
Looks good, maybe fix indentation

nikriek (approved):

Issue #91, opened 2017-11-20
Translate the current application


PoJo93:
**AS** User
**I WANT TO** have every menu Item in German 
**TO** see the application in German and be able to use it as a German speaker
**ACCEPTANCE CRITERIA**
- [x] the menu should be in German
- [x] the columns in player should be in German
- [x] the columns in teams should be in German
- [x] the columns in matches should be in German
- [x] the buttons in matches should be in German
- [ ] my profile page should be in German
- [x] the labels of my profile page should be in German

manner:
translate the validation of photo upload

manner:
original size: 3

PoJo93:
When system language is English than the picture upload is also in English but should be German

PoJo93:
Still the error with the picture upload persists in my system. Was there a problem to always set the language for the picture upload to german @manner

PR #416, opened 2018-01-22
Set default german locale. ref #91.

manner:


nikriek (approved):
You might also be able to hardcode the locale in an initializer. This reduces the LOC to 1 ;)

manner:
[Rails Guide](http://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests) recommends setting I18n locale in Application Controller

PR #418, opened 2018-01-22
345/bruteforce/event not deleteable after start

### LINKED ISSUE #345: events should not be deleteable after their start date
It should not be possible to delete an event after it started, because otherwise progress will be lost.
After its start an event should only be finished (https://github.com/hpi-swt2/sport-portal/issues/233) and archived.

**Acceptance Criteria**

- [ ] an event can only be deleted until the event started
- [ ] the admin can delete it any time

usame-algan:
closes #345

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -28,7 +28,7 @@ class Event < ApplicationRecord
   has_many :editors, through: :organizers, source: 'user'
 
   scope :active, -> { where('deadline >= ? OR type = ?', Date.current, "Rankinglist") }
-  scope :started, -> { where('startdate >= ? OR type = ?', Date.current, "Rankinglist") }
+  scope :not_started_yet, -> { where('startdate >= ? OR type = ?', Date.current, "Rankinglist") }
 
   enum selection_type: [:fcfs, :fcfs_queue, :selection]
   validates :name, :discipline, :game_mode, :player_type,  presence: true
@@ -54,10 +54,6 @@ class Event < ApplicationRecord
 
   def deadline_has_passed?
     deadline < Date.current
-  end
-
-  def startdate_has_passed?
-    startdate < Date.current
   end
 
   def add_team(team)
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -55,6 +55,10 @@ class Event < ApplicationRecord
     deadline < Date.current
   end
 
+  def startdate_has_passed?
+    startdate < Date.current
+  end
+
   def add_team(team)
     teams << team
     invalidate_schedule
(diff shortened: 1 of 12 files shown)


spec/features/event/show_event_spec.rb
```diff
@@ -203,6 +203,18 @@
...
+        @ongoing_event = FactoryBot.create :event, startdate: Date.yesterday
+        sign_in @user
+        visit event_path(@ongoing_event)
+      end
+
+      it "should not display a delete button" do
+        expect(page).not_to have_css('a', :text => I18n.t('helpers.links.destroy'))
+      end
+    end
+
```
Askir: If we're working with cucumber now this test isn't needed anymore. Cucumber replaces all our feature tests.

spec/models/ability_spec.rb
```diff
@@ -145,7 +145,7 @@
   end
 
   it 'should allow users to crud events they created' do
-    event = Event.new(owner: @user)
+    event = FactoryBot.create(:event, :has_dates, owner_id: @user.id)
```
Askir: This is awkward because it's not really obvious why the even needs to have dates, I'd rather see this in a sperate ability test only for Tournaments and leagues, because rankinglists don't have any dates. Also adding another test with the other case (when you cannot delete the event) would be a good idea imo.

features/events/delete_event.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  Scenario: Admin and organizer can delete event that has not yet started
+    Given an event that has not started yet
+    Then the admin should be able to delete it
+    And the organizer should be able to delete it
+
+  Scenario: Admin can delete event any time, organizer does not
+    Given an event that has started
+    Then the admin should be able to delete it
+    And the organizer should not be able to delete it
```
Askir: Implement the steps for this please :) If you need help don't be afraid to ask, although my knowledge about cucumber is also still quite limited.

app/models/event.rb
```diff
@@ -55,6 +56,10 @@ def deadline_has_passed?
     deadline < Date.current
   end
 
+  def startdate_has_passed?
+    startdate < Date.current
+  end
```
Askir: This method is only used in your tests, nowhere else. Is that intentional? I don't really see how this adds the expected behaviour yet.

Askir (changes requested):

spec/models/ability_spec.rb
usame-algan: Users can only delete events now under certain conditions so this test has to be changed. But I agree that the event object having a date here is weird. I will change this test and create a separate one for the changed delete ability.

app/models/event.rb
usame-algan: This is from an earlier implementation that is actually not needed anymore.

Askir (approved):

Askir:
Can you merge this with dev, so we can push it into PO Review? :)

Issue #367, opened 2018-01-17
Extract ranking logic into dedicated class


birneamstiel:
Make the `EventsController` slim again by moving logic calculating the ranking logic into new Class (something like `RankingEntry` and `RankingList`). The `Event` model would probably have a `#get_ranking` method that would calculate the ranking list.

- follow-up for #125
- see https://github.com/hpi-swt2/sport-portal/pull/286#discussion_r161774881

erksch:
Whats the status of this? I need the extracted ranking method to implement the swiss system.

birneamstiel:
@erksch I'd say you have a look at the PR referenced above :-)

https://github.com/hpi-swt2/sport-portal/pull/419

birneamstiel:
needed time: 4h

PR #419, opened 2018-01-22
Refactor ranking action

birneamstiel:
... by extracting logic from `EventsController` into two new classes (`RankingList` and `RankingEntry`) as suggested by @Askir (see https://github.com/hpi-swt2/sport-portal/pull/286#discussion_r161774881)

@stlutz @Jeilef Conrad pointed out you are working on event rankings, too. That's why I've requested reviews from you guys. I think the whole refactoring is already in a pretty final state, so you can start working with this new structure (some things might change though, see comment below).

Issue: #367


birneamstiel:
@reviewers 
### things to discuss:
- position of new classes in project directory structure:
  - I've decided to put those into the `model` folder, but there are other options:
  - see [here](http://codesthq.com/blog/2015/hi-im-poro.html) and [here](http://vrybas.github.io/blog/2014/08/15/a-way-to-organize-poros-in-rails/)
- do we want to change the structure of tests that are relevant to the ranking logic, too?
  - lot of tests are in the `events_controller_spec.rb` file, which I think is fine, since the ranking calculation is still somehow triggered by the `EventsController`
  - but one could also argue to move those tests into test files testing only the newly created classes
- do we need the `RankingList` class?
  - it might be better to move the whole (rather slim) logic contained it `RankingList`'s `get_ranking` method into the `Event` model
- maybe `RankingList` is not the best name for something that's creating an array? 
  - on the other hand it's pretty descriptive...

app/models/ranking_entry.rb
```diff
@@ -0,0 +1,69 @@
+class RankingEntry
+  attr_reader :rank, :name, :match_count, :won_count, :draw_count, :lost_count, :goals, :goals_against, :goals_difference, :points
+  attr_accessor :rank
```
birneamstiel: @reviewers does anybody know what's the nicest way to deal with accessible attributes in ruby? Code climate gives a warning, might be better to create a accessor method? Or is this done internally?

app/controllers/events_controller.rb
```diff
@@ -100,77 +100,7 @@ def schedule
...
-    ranking_entry.goals += score_home
-    ranking_entry.goals_against += score_away
-    ranking_entry.points += match.points_home
-  end
-
-  def parse_match_details_for_away(ranking_entry, score_home, score_away, match)
-    ranking_entry.goals += score_away
-    ranking_entry.goals_against += score_home
-    ranking_entry.points += match.points_away
+    @ranking_entries = @event.get_ranking
```
Askir: lovely

app/models/ranking_entry.rb
Askir: I'm pretty sure this is exactly what `attr_reader` and `attr_accessor` do. Not sure why codeclimates gives a warning tbh.

app/models/ranking_list.rb
```diff
@@ -0,0 +1,31 @@
+# Contains logic for calculating an array containing ranking entries out of an array of teams and matches.
+class RankingList
```
Askir: The name will clash a little bit with the rankinglist model we already have. Maybe something shorter like just Ranking is also precise enough? There should definitely be a renaming here in my opinion.

Askir (commented):
Didn't actually look at the logic or anything, I'll just trust you on that :P But it looks really nice. One thing: The POROs don't really belong in the models folder. In my experiences they are usually placed in /lib.

app/models/ranking_list.rb
birneamstiel: `Ranking` would be fine from my point of view, @bakoe @jonasnoki what do you say?

app/models/ranking_entry.rb
```diff
@@ -0,0 +1,70 @@
+# Smallest unit of a ranking list. Contains different data for one team and contains logic to parse those date out of
```
bakoe: Typo in `date`:
```diff
- … those date out of
+ … those data out of
```

Depending on your preference, you could also change `those` to `these` – see https://www.englisch-hilfen.de/grammar/this_that.htm:

> Ist etwas näher dran, nimmt man this/these, ist etwas weiter weg (zeitlich und örtlich) steht that/those.

app/models/ranking_entry.rb
```diff
@@ -0,0 +1,70 @@
...
+# match objects.
+class RankingEntry
+  attr_reader :rank, :name, :match_count, :won_count, :draw_count, :lost_count, :goals, :goals_against, :goals_difference, :points
+  attr_accessor :rank
```
bakoe: To my mind, the `attr_accessor :rank` makes the `attr_reader :rank` redundant since `attr_accessor` includes an attribute reader as well as a writer. See http://ruby-doc.org/core-2.0.0/Module.html#method-i-attr_accessor

app/models/ranking_list.rb
bakoe: I agree: `Ranking` sounds reasonable. 👍

bakoe: If you change the name of this class, make sure to change the explanatory comment from `ranking_entry.rb`'s first line as well.

bakoe (commented):
LGTM code-wise! Only some minor comments regarding naming, redundancy and a typo.

bakoe:
@birneamstiel Regarding the *things to discuss*:

> - position of new classes in project directory structure:
>   - I've decided to put those into the `model` folder, but there are other options:
>   - see [here](http://codesthq.com/blog/2015/hi-im-poro.html) and [here](http://vrybas.github.io/blog/2014/08/15/a-way-to-organize-poros-in-rails/)

I personally like the namespace-based approach taken from http://vrybas.github.io/blog/2014/08/15/a-way-to-organize-poros-in-rails/ (see _Example with Models_) instead of putting the POROs in `model` or `lib`. 
> - do we want to change the structure of tests that are relevant to the ranking logic, too?
>   - lot of tests are in the `events_controller_spec.rb` file, which I think is fine, since the ranking calculation is still somehow triggered by the `EventsController`
>   - but one could also argue to move those tests into test files testing only the newly created classes

I think the tests can stay in the EventsController for the moment … this allows refactoring the `Event`, `Ranking` and `RankingEntry` classes without making the existing tests fail. 

> - do we need the `RankingList` class?
>   - it might be better to move the whole (rather slim) logic contained it `RankingList`'s `get_ranking` method into the `Event` model

To my mind, it's fine to have this as a separate class. :+1:

> - maybe `RankingList` is not the best name for something that's creating an array? 
>   - on the other hand it's pretty descriptive...

Regarding the comments from @Askir, `Ranking` would probably be a better name anyway because it causes less confusion regarding the already existing `RankingList`.

lheilmann (approved):
LGTM all in all! I like the namespace approach - Namespaces FTW! I did not recognize any changes in the ranking logic itself. I reviewed the logic already, so to my mind we are done here! :+1:

jonasnoki:
image
Works great!

Issue #343, opened 2018-01-15
Participation Type should be mandatory


MarcelSchmidberger:
The "Teilnahmeform" field should be mandatory.
Right now it is not necessary to be filled in but when creating the event will be autofilled with "Einzelspieler". It is mandatory when editing the event but not when creating it. This is inconsistent.  This field should just be mandatory everywhere to prevent confusion.

PR #421, opened 2018-01-22
#343/bruteforce/mandatory participation type

laurenzse:


Jeilef (commented):
surprisingly unspectacular for how many Reviewers you added

app/controllers/events_controller.rb
```diff
@@ -184,8 +184,8 @@ def set_event
 
     def set_associations
       @event.owner = current_user
-      @event.player_type ||= Event.player_types[:single]
-      if @event.player_type == Event.player_types[:single] || @event.type == 'Rankinglist'
+      if @event.type == 'Rankinglist'
```
Askir: I think instead of putting this quite detailed logic into the container, you can use a before_save hook in the model, that ensure the playertype is set to single before saving a rankinglist. (Pretty sure that is what this line is for, correct?) it would also avoid the rather ugly type checking.

spec/views/events/form_examples.rb
```diff
@@ -34,6 +34,24 @@
...
+      expect(rendered).to have_field(Event.human_attribute_name :player_type)
+    end
+
+    it 'has default input for player_type field' do
+      render
+
+      if path == :edit
+        expect(rendered).to have_field(Event.human_attribute_name :player_type, with: t(Event.human_player_type(@event.player_type)))
+      elsif path == :new
+        expect(rendered).to have_field(Event.human_attribute_name :player_type, with: t('events.new.select_player_type'))
```
Askir, line 50: not a huge fan of splitting it up like this, the `with:` is omitted for all other tests. But I do see the value of specifically testing the default input, since it is exactly what this issue is about. We should just consider splitting the :new and :edit tests into two different sharde_examples at some point, since otherwise this one will look like one huge switch case.
Fine for now though.

spec/views/events/new.html.erb_spec.rb
```diff
@@ -8,15 +8,15 @@
       assign(:event, FactoryBot.build(:league))
     end
 
-    it_should_behave_like 'an event form', for_class: League, with: [:dates, :capacity, :gameday_duration, :selection]
+    it_should_behave_like 'an event form', for_class: League, with: [:dates, :capacity, :gameday_duration, :selection, :player_type]
```
Askir, line 11: At least the reference of the test looks pretty clean I still like that ;)

Askir (changes requested):
Try to keep the controllers and models as lean as possible, if you could extract those two lines and get rid of the type check: looks good to me :)

Askir (approved):

PR #425, opened 2018-01-23
Bruteforce/#128/playing schedule with link to players

laurenzse:


spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -2,23 +2,52 @@
...
-    assign(:event, @league)
-    assign(:schedule_type, 'league')
-    assign(:matches, @league.matches)
-  end
+  describe 'page layout' do
+    before :each do
+      @league = FactoryBot.create(:league, :with_matches,
+                                  deadline: Date.parse('23.12.2017'),
+                                  startdate: Date.parse('24.12.2017'),
+                                  enddate: Date.parse('31.12.2017'),
```
usame-algan, line 10: I think you can use the `has_dates` trait here.

spec/features/event/schedule_league_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    @league.add_participant(@user)
+    @league.add_participant(@another_user)
+  end
+
+  it 'has link to player' do
+    visit event_schedule_path(@league)
+
+    click_on(@user.name)
+    expect(page).to have_current_path(user_path(@user))
```
usame-algan, line 18: This looks good 👍

spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -2,23 +2,52 @@
...
+                                  gameday_duration: 7)
+      assign(:event, @league)
+      assign(:schedule_type, 'league')
+      assign(:matches, @league.matches)
```
usame-algan, line 14: Are these variables used anywhere? `event, schedule_type, matches`

usame-algan (changes requested):

spec/views/events/schedule.html.erb_spec.rb
laurenzse: A test refers to those exact dates (for rendering purposes). I think using the has_dates trait would make things more complicated there.

spec/views/events/schedule.html.erb_spec.rb
laurenzse: I didn't write those lines, but removing them makes the tests red. Maybe they are used within rendering?

Askir: The assign statements replace the controller lines:
```Ruby
@event = @league
@schedule.type = 'league'
@matches = league.matches
```
Since the controller function isn't called in view tests you have to assign these values manually to render the template. Was that the question? @usame-algan

spec/views/events/schedule.html.erb_spec.rb
Askir: Since these specific dates are important for the test. I'd rather specify them here. If we don't do that and somebody changes the dates in `:has_dates` to someting else. This test will break.
So my point is: It's not only important that our league has dates but it's also important what dates it actually has.

spec/views/events/schedule.html.erb_spec.rb
usame-algan: Yes, I was wondering what those lines were for since I didnt find them used anywhere in the file. Thanks for clarification!

usame-algan (approved):

PR #430, opened 2018-01-24
Fix start_time display in Spielplan 

nikriek:


Askir (approved):
not the most beautiful way to fix this but better than not having a schedule at all.

Issue #183, opened 2017-11-29
Header design


stursberg:
Header will be shown on the top of every page. There is no top-margin needed.
  
![Screen Shot 2017-11-29 at 16.36.21.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/59fc457019d4e4001305101f/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76226456fa6b7191a379e90c2857415d4fbe5d4a1f06a3a3dc78e2a436628f0d0beed23f0b26933a43a9b8841fbf02b325df1e87e7ba9850d2aa6c7bb3d4d216744b470a83ed5c32aa984ec959ff4de84555a2113f63f43b.png)
 
- Maximum width of content (= header from logo to login button): 1200px; Width of background (color/pic) is not limited;
- transitions when menu opens (open from top to bottom), use materialize again?
- all data (for example font size, colors etc - in mock; ask Dima for CSS);
- Font: ProximaNova

**Estimation: XL**

Version for mobile (just styling, no new menu entries, 2nd pic --> click on Dmitri --> 3rd pic)
![Screen Shot 2017-11-29 at 16.51.03.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/59fc457019d4e4001305101f/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76226456fa6b7191a379e90c2e50415f4dbe5d4a1f06a3a3dc78e2a436628f0d0beed23f0b26933a43a9b8841fbf02b325df1e87e7ba9850d2aa6c7bb3d4d216744b470a83ed5c32aa984ec959ff4ce14b52ad113764f43b.png)

The result should look like this (Acceptance Criteria):
 
![Screen Shot 2017-11-29 at 17.05.13.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/59fc457019d4e4001305101f/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76226456fa6b7191a379e80c2b54415e4dbe5d4a1f06a3a3dc78e2a436628f0d0beed23f0b26933a43a9b8841fbf02b325df1e87e7ba9850d2aa6c7bb3d4d216744b470a83ed5c32aa984ec959ff4ce04a51a6163f67f43b.png)

- [ ] Implement styling of desktop view
- [ ] Implement styling of mobile view

stursberg:
[...]

PR #432, opened 2018-01-25
Flexibus/#379/teams page design

### LINKED ISSUE #379: Teams view as tiles

emanuel-metzenthin:
Integrated requested changes, closes #379.


app/assets/stylesheets/navbar.css
```diff
@@ -169,6 +169,7 @@ img.avatar {
 
 .container, .container-fluid{
     max-width: 1200px;
+    width: 100% !important;
```
deeps96, line 172: Why is this change assigned to this ticket? - I think it belongs to #183.

deeps96 (changes requested):
Remove this selection:
image

There is still some intend, which was explicitly not requested by PO. @zhamanakov Is this ok?

image

Intendation is slightly bigger on square images.

Overall code looks okay to me.

app/assets/stylesheets/navbar.css
emanuel-metzenthin: That was part of Dimas requested changes for the event tiles as bootstrap somehow always set the container width to 70%.

deeps96: FYI: The container stuff is generally wanted and comes from Materialize.

deeps96 (approved):

PR #433, opened 2018-01-26
Make private teams on teams index page invisible for not-logged in or non-member users.

jonasnoki:


app/controllers/teams_controller.rb
```diff
@@ -1,15 +1,15 @@
...
   before_action :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
-  authorize_resource :team, only: [:index, :show, :new, :edit, :create, :update, :destroy]
+  load_and_authorize_resource :team, only: [:index, :show, :new, :edit, :create, :update, :destroy]
 
   # GET /teams
   def index
     multi_teams = Team.multiplayer
     if params[:filter] == "true"
-      @teams = multi_teams.includes(:team_members).where(team_users: { user_id: current_user })
+      @teams = @teams - (@teams - multi_teams.includes(:team_members).where(team_users: { user_id: current_user }))
```
lheilmann: @jonasnoki Could you please explain, why this expression has to be like that?

spec/controllers/teams_controller_spec.rb
```diff
@@ -42,6 +46,31 @@
...
+    teams = controller.instance_variable_get(:@teams)
+    expect(teams.first).to eq(team)
+  end
+
+  it "should not show  private teams to not-logged in user" do
+    team = Team.create! private_attributes
+    team.owners << @user
+    sign_out @user
+    get :index, params: {}
+    teams = controller.instance_variable_get(:@teams)
```
lheilmann: That's cool!

spec/controllers/teams_controller_spec.rb
```diff
@@ -42,6 +46,31 @@
...
+    expect(teams).to be_empty
+  end
+
+  it "should not show  private teams to a user who is not part of the team" do
+    team = Team.create! private_attributes
+    team.owners << @other_user
```
lheilmann: I would assume you do not need this line of code, because the factory creates a team owner after creating the team. See https://github.com/hpi-swt2/sport-portal/blob/dev/spec/factories/teams.rb.

spec/controllers/teams_controller_spec.rb
```diff
@@ -42,6 +46,31 @@
...
+    team = Team.create! private_attributes
+    team.owners << @user
+    get :index, params: {}
+    teams = controller.instance_variable_get(:@teams)
+    expect(teams.first).to eq(team)
+  end
+
+  it "should not show  private teams to not-logged in user" do
+    team = Team.create! private_attributes
+    team.owners << @user
```
lheilmann: See comment below. (Line 69)

lheilmann (changes requested):
Only minor changes. LGTM! :+1:

app/controllers/teams_controller.rb
bakoe: @jonasnoki I guess you just want to make `@teams` the intersection of (the `load_and_authorize`d) `@teams` and `multi_teams.includes(…)`? You could just try using the `&` operand of Ruby‘s Arrays then which should do the same.

lheilmann (approved):

app/controllers/teams_controller.rb
jonasnoki: @bakoe got it right. & operator works like a charm.

spec/controllers/teams_controller_spec.rb
jonasnoki: It's not inserting the @other_user though. Might not need it in this place makes it clearer though.

spec/controllers/teams_controller_spec.rb
jonasnoki: thx

PR #435, opened 2018-01-26
Enable changing dates for gamedays

Askir:
This is almost done. I just need somebody to teach me enough css to not make it look ridiculous @usame-algan / @erksch .
Gameday dates don't update automatically yet. Shouldn't be too difficult to add later though.

Changes aren't that huge actually but I renamed a column in the matches table so I had to touch a lot of files :D


Askir:
I hope the implementation of the gameday model is fine. I know that `gameday_number` as a rename is probably not ideal but I think using the gameday model in the future will be the better choice anyways so I hope that we can remove that column eventually (slow refactoring while we are still progressing with the product like we learned just before christmas :D).

app/controllers/gamedays_controller.rb
```diff
@@ -0,0 +1,11 @@
+class GamedaysController < ApplicationController
+  def update
+    @gameday = Gameday.find(params[:id])
+    @gameday.update(gameday_params)
+    redirect_to event_schedule_path(@gameday.event), notice: 'successfully changed gameday date'
```
erksch: The `notice` should be translatable.

app/models/ability.rb
```diff
@@ -84,6 +85,12 @@ def can_leave_event(user)
       end
     end
 
+    def can_update_gameday(user)
+      can :update, Gameday do |gameday|
+        not (gameday.event.organizers & user.organizers).empty?
```
erksch, line 90: Why `user.organizers`?

Askir: because organizer is a join between even and user. It belongs to one user and one event. And because there is no `has_many: users, through: organizers` in Event and no `has_many: events, through: organizers` in User. I have to compare them like this.

Askir: Mabye for clarification: `&` is the cut volume of two arrays.

app/controllers/gamedays_controller.rb
Askir: True, fixed it :)

erksch (approved):

PR #440, opened 2018-01-27
 Swteam/#57/ mail confirmations

### LINKED ISSUE #57: Mail verification: confirmation Emails

Markus-Brand:
Users now need to confirm their email address on signup. They also need to reconfirm a new email address when changing their email.

Will close #57


config/initializers/devise.rb
```diff
@@ -125,7 +125,9 @@
   # able to access the website for two days without confirming their account,
   # access will be blocked just in the third day. Default is 0.days, meaning
   # the user cannot access the website without confirming their account.
-  # config.allow_unconfirmed_access_for = 2.days
+
+  # a value bigger than 0 is required here for all our tests to work
```
fpottbaecker, line 129: Why?

db/migrate/20180126101234_add_unconfirmed_email.rb
```diff
@@ -0,0 +1,16 @@
+# as seen here: https://github.com/plataformatec/devise/wiki/How-To:-Add-:confirmable-to-Users
+class AddUnconfirmedEmail < ActiveRecord::Migration[5.1]
+  def up
+    add_column :users, :confirmation_token, :string
```
fpottbaecker: Could you use a change_table here instead? That way you can specify the column type via the method name [1](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_table).

db/migrate/20180126101234_add_unconfirmed_email.rb
```diff
@@ -0,0 +1,16 @@
...
+    add_column :users, :confirmed_at, :datetime
+    add_column :users, :confirmation_sent_at, :datetime
+    add_column :users, :unconfirmed_email, :string
+    add_index :users, :confirmation_token, unique: true
+    User.all.update_all confirmed_at: DateTime.now
+  end
+
+  def down
```
fpottbaecker: I think you can make this migration shorter by using `reversible` and only specifying up with update all. That way you can use the `change` method.

features/step_definitions/user.rb
```diff
@@ -209,7 +209,14 @@
   expect(single_user.email).to eq(email)
 end
 
-
+And(/^he clicks the link in the email he just received$/) do
+  # as seen here: https://stackoverflow.com/questions/25240102/how-to-visit-a-link-inside-an-email-using-capybara
```
fpottbaecker: Didn't we use capybara-email in an earlier test?

spec/controllers/users_controller_spec.rb
```diff
@@ -250,6 +250,56 @@
...
+        old_mail = user.email
+        new_mail = 'my_new_mail@example.com'
+        new_attributes = { email: new_mail,
+                           current_password: user.password }
+        put :update, params: { id: user.to_param, user: new_attributes }
+
+        expect(ActionMailer::Base.deliveries.length).to eq(2)
+        notification_mail = ActionMailer::Base.deliveries.first
+        notification_mail_body = Capybara.string(notification_mail.body.encoded)
+        expect(notification_mail.to.length).to eq(1) #one receiver
```
fpottbaecker: I think this should be part of a mailer spec, you should only verify that the right mailer is invoked.

spec/controllers/users_controller_spec.rb
```diff
@@ -250,6 +250,56 @@
...
+
+        confirmation_mail = ActionMailer::Base.deliveries.second
+        confirmation_mail_body = Capybara.string(confirmation_mail.body.encoded)
+        expect(confirmation_mail.to.length).to eq(1) #one receiver
+        expect(confirmation_mail.to.first).to eq(new_mail) #that is our users new mail address
+        expect(confirmation_mail.subject).to eq(I18n.t('devise.mailer.confirmation_instructions.subject'))
+        expect(confirmation_mail_body).to have_link(I18n.t('devise.mailer.confirmation_instructions.action'))
+      end
+
+      it 'should receive a confirmation mail on sign-up' do
```
fpottbaecker: This looks more like a feature spec. Also, it's not the controller who "should receive a confirmation mail on sign-up".

spec/controllers/users_controller_spec.rb
```diff
@@ -250,6 +250,56 @@
...
+        expect(ActionMailer::Base.deliveries.length).to eq(1)
+        confirmation_mail = ActionMailer::Base.deliveries.first
+        confirmation_mail_body = Capybara.string(confirmation_mail.body.encoded)
+        expect(confirmation_mail.to.length).to eq(1) #one receiver
+        expect(confirmation_mail.to.first).to eq(user.email) #that is our user
+        expect(confirmation_mail.subject).to eq(I18n.t('devise.mailer.confirmation_instructions.subject'))
+        expect(confirmation_mail_body).to have_link(I18n.t('devise.mailer.confirmation_instructions.action'))
+      end
+
+      it 'should receive no confirmation mail when signing up via omniauth' do
```
fpottbaecker: same here

fpottbaecker (changes requested):

config/initializers/devise.rb
ErikBrendel: because otherwise the accounts are not ready to use. You would have to manually click on the verification link / disable this verification for all the tests, for it to work

db/migrate/20180126101234_add_unconfirmed_email.rb
ErikBrendel: Yes I could. I see no reason why I should. You can do if you want to

db/migrate/20180126101234_add_unconfirmed_email.rb
ErikBrendel: Yes, I could, but I see no reason. Copying it from the webpage referenced worked aswell. But I will not prevent you from tidying this up :D

config/initializers/devise.rb
fpottbaecker: Couldn't we add that as default to the factory with a trait `unconfirmed`?

spec/mailers/user_mailer_spec.rb
```diff
@@ -1,11 +1,71 @@
 require 'rails_helper'
 
-RSpec.describe DeviseMailer, type: :mailer do
-  describe 'confirmation mails' do
+RSpec.describe Devise.mailer, type: :mailer do
+
```
fpottbaecker: Space — the final frontier. These are the voyages of the starship `develop`. Its ongoing mission — to program strange new apps — to seek out new smells and new optimizations — to boldly code where no man has coded before.

spec/mailers/user_mailer_spec.rb
```diff
@@ -1,11 +1,71 @@
...
+
+  describe 'confirmation mail' do
+    let(:user) { FactoryBot.create :user, unconfirmed_email: 'my_new_email@example.com' }
```
fpottbaecker: Can you move this let out of both blocks?

config/initializers/devise.rb
```diff
@@ -125,7 +125,9 @@
...
+  config.allow_unconfirmed_access_for = 1.day
```
Oluwoye, line 130: Gets the user sufficient information about that time limit?

spec/controllers/users_controller_spec.rb
```diff
@@ -250,6 +264,27 @@
...
+        user = User.create! valid_attributes
+        sign_in user
+        Devise.mailer.deliveries.clear
+
+        new_mail = 'my_new_mail@example.com'
+        new_attributes = { email: new_mail,
+                           current_password: user.password }
+
+        mock_confirmation_mail = double
+        mock_notification_mail = double
```
Oluwoye: Why do you double this line?

spec/controllers/users_controller_spec.rb
```diff
@@ -250,6 +264,27 @@
...
+        expect(Devise.mailer).to receive(:confirmation_instructions).and_return(mock_confirmation_mail)
+        expect(Devise.mailer).to receive(:email_changed).and_return(mock_notification_mail)
+        expect(mock_confirmation_mail).to receive(:deliver)
+        expect(mock_notification_mail).to receive(:deliver)
```
Oluwoye: This is doubled too

spec/models/match_spec.rb
```diff
@@ -186,9 +186,11 @@
   end
 
   it 'send notification emails to all members of teams when scheduled' do
-    match = FactoryBot.create(:match)
-    email_count = match.team_home.members.count + match.team_away.members.count
-    expect { FactoryBot.create(:match) }.to change { ActionMailer::Base.deliveries.length }.by(email_count * 2)
+    #match = FactoryBot.build(:match)
```
Oluwoye: Remove this line, if you don't need it

Oluwoye (changes requested):

spec/controllers/users_controller_spec.rb
```diff
@@ -167,6 +167,20 @@
         post :create, params: { user: valid_attributes }
       }.to change(User, :count).by(1)
     end
+
+    it 'sends a mail confirmation mail to new users' do
+      @request.env["devise.mapping"] = Devise.mappings[:user]
```
nikriek: You should create an rspec module/helper for
`@request.env["devise.mapping"] = Devise.mappings[:user]`

nikriek (changes requested):

spec/controllers/users_controller_spec.rb
fpottbaecker, line 270: This is intentional, there should be two separate emails that are both delivered.

Oluwoye: Ok^^

config/initializers/devise.rb
ErikBrendel: would also be possible. On the other hand, (if i remember correctly) the PO did not dislike this setting in general. So we can leave it as it is anyways, and there is imho no need to adjust the tests so they would work without it

spec/controllers/users_controller_spec.rb
ErikBrendel: there is the `notification` and the `confirmation` mail, and both should get send

spec/controllers/users_controller_spec.rb
ErikBrendel: how exactly would that look like? Could you provide me a link or do it yourself on this branch?

config/initializers/devise.rb
ErikBrendel: Interesting question :D

I just tested it: You do not see this time limit anywhere. But you immediately get the account confirmation mail. When you try to sign in after the day has passed, you get the error message "Du musst Deinen Account bestätigen, bevor Du fortfahren kannst."

Is this enough?

Oluwoye: Hmm, I think, it is more convenient, if the error message is something like "Der genutzte Link ist leider abgelaufen. Fordere einen neuen an."

spec/controllers/users_controller_spec.rb
nikriek: https://github.com/plataformatec/devise/wiki/How-To:-Test-controllers-with-Rails-3-and-4-%28and-RSpec%29

spec/mailers/user_mailer_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(mail.from).to eq(['sport.portal@gmx.de'])
+    end
+
+    it 'assigns the user\'s firstname' do
+      expect(mail.body.encoded).to match(user.first_name.capitalize)
+    end
+
+    it 'has a confirmation link' do
+      mail_body = Capybara.string(mail.body.encoded)
+      expect(mail_body).to have_link(I18n.t('devise.mailer.confirmation_instructions.action'))
```
fpottbaecker: Could you also test the `href` here?

spec/factories/users.rb
```diff
@@ -36,8 +36,9 @@
     sequence(:telegram_username) { |n| "telegram_user#{n}" }
     favourite_sports "Football, Basketball, Tennis"
 
-    factory :invalid_user do
-      last_name nil
+
+    after(:create) do |user|
+      puts 'foo'
```
fpottbaecker: Ummm... no?

config/initializers/devise.rb
fpottbaecker: This is not the amount of time you have to finally confirm your account. This is the amount of time you can use your account without confirming.

spec/controllers/users_controller_spec.rb
fpottbaecker: Fixed by removing unnecessary calls. This is already done `before :each` in this test, so no duplication.

config/initializers/devise.rb
Oluwoye: Then it is ok

Oluwoye (approved):

fpottbaecker (approved):

spec/support/controller_macros.rb
```diff
@@ -0,0 +1,5 @@
+module ControllerMacros
+  def mock_devise
+    @request.env['devise.mapping'] = Devise.mappings[:user]
```
nikriek: You may want to neste this in `before(:each)` as described in https://github.com/plataformatec/devise/wiki/How-To:-Test-controllers-with-Rails-3-and-4-%28and-RSpec%29. Otherwise looks good

nikriek (approved):

Issue #272, opened 2017-12-19
Results of ranking lists


GittiHab:
**AS A** participant of a ranking list event
**I WANT TO** be able to add matches to ranking lists
**SO THAT** the score can be displayed

## ACCEPTANCE CRITERIA
- [x] On the detailed page of a ranking list event there should be a button "Add match" next to each participant in the participants/ranking list if the logged in user is a participant him-/herself.
- [x] When clicking the button the user should be shown a form where he can specify the result of the match against the other player (next to whom the button was)
- [x] When the user saves the form the match should be saved and both users score (the user who added the match and the opponent) should be updated according to the metric of the ranking list

GittiHab:
This user story needs feedback: Should the opponent be able to verify the result first, how does this fit in to the existing match-event system?

GittiHab:
This story is linked to #135

GittiHab:
I updated the acceptance criteria according to the discussion with @Jeilef.

GittiHab:
Looks good

PR #442, opened 2018-01-28
Issuenumber5/#135/result list of ranking lists

pawelsoundso:
Adds a list of participants of ranking lists, with an initial value on join and a button, which should be modified later by #272.


app/views/events/show.html.erb
```diff
@@ -82,10 +82,27 @@
...
+    </tr>
+    </thead>
     <tbody>
-    <% @event.teams.each do |participant| %>
-      <tr><td><%= link_to_participant participant %></td></tr>
+    <% @event.participants.order(rating: :desc).each do |participant| %>
+      <tr><td><%= link_to_participant participant.team %></td>
+        <% if @event.is_a? Rankinglist %>
+          <td><%= participant.rating %></td>
+          <% if can? :leave, @event %>
```
sumpfranze: is leave the correct action to test whether a match can be added or not?

sumpfranze (commented):
otherwise lgtm

app/views/events/show.html.erb
Jeilef: this is implemented in #272

Jeilef (approved):
like Jacob, Looks Good To Merge

Issue #224, opened 2017-12-09
Example data should be automatically loaded


MarcelSchmidberger:
As a Tester of the platform,
I want to see realistic Users, Teams and Events,
I oder to test features without the need to create everything from scratch

**Acceptance Criteria**

- [ ] realistic data is always in the platform
- [ ] There should be at least 12 Users, 6 Teams with each 2 Users and 4 Events

nikriek:
Consider removing the data in `db/seeds.rb` such as: https://github.com/hpi-swt2/sport-portal/issues/234

PR #445, opened 2018-01-28
Add fake users to seed (closes #234)

### LINKED ISSUE #234: Create Sample User Data

nikriek:
Connects #234


Gemfile
```diff
@@ -101,6 +101,8 @@ gem 'awesome_nested_fields'
 # A modern, responsive front-end-framework based on Material Design
 gem 'materialize-sass'
 
+gem 'faker'
```
Askir: please add a short description of what this gem does and preferably a repository link for further documentation

db/seeds.rb
```diff
@@ -55,6 +56,24 @@
...
+    last_name: Faker::Name.unique.last_name,
+    email: Faker::Internet.unique.email,
+    password: 'seed1234',
+    password_confirmation: 'seed1234',
+    birthday: Faker::Date.birthday(18, 65),
+    telegram_username: 'eb',
+    telephone_number: '12345654321',
+    avatar: File.open("#{Rails.root}/db/seed_data/avatar3.png"),
+    favourite_sports: sports[rand(sports.length), rand(sports.length - 1) + 1].join(', ')).save(validate: false)
+end
```
Askir, line 75: It's not really the job of the seeds file to add example data. The seeds file should only specify the minimal data that is needed for operating the system at all.

But honestly I think for now this is fine. Just consider opening a ticket to extract this behaviour into a sperate rake task. We also have another ticket for even more example data that you could link to: #224

Gemfile
nikriek: Done

db/seeds.rb
nikriek: Jep, you are right. I added a comment to the ticket that you mentioned.

Lariiii (approved):

PR #449, opened 2018-01-28
Issuenumber5/#221/min max players in teams. Work in PO review

fswt:


spec/models/ability_spec.rb
```diff
@@ -250,4 +249,52 @@
       include_examples "a single player event"
     end
   end
+
```
ConstantinLange, line 287: You can put all the shared code of the tests in a "before" block. That safes a lot of space.

Jeilef: line 72 - 156

Jeilef (approved):
much wow

PR #437, opened 2018-01-26
Notify users by email about being added to a team and about their team participating in an event

bakoe:
see #437

Issue #301, opened 2018-01-08
Notifications for user: email settings


svitlanakalancha:
As _User_ I want to be _informed_ about changes in my _events, teams or event calendar_ to plan my participation accordingly. 

**Acceptance criteria**

- Notification is an automated email sent to the mailbox of the user
- Only registered users can receive notifications
- Emails can be turned on/off on the page "Benachrichtigungseinstellungen". By default notifications are turned on. 

![pasted_image_17_01_18__14_23](https://user-images.githubusercontent.com/27727717/35044824-0f84d2e0-fb92-11e7-88cb-3cd7b1e070f6.png)

**Important**
Check issues #57 and #149 and communicate with other dev teams

**Out of scope**
Implementation of notifications on the website should not be done.

lheilmann:
@svitlanakalancha Size of "Send mails in rails applications" (Spike):  5

lheilmann:
@svitlanakalancha Size of "Research on jobs on rails server targeting event emails - see #301" (Spike): 5

lheilmann:
Checkbox "Newsletter abonnieren" is out of scope. Cheers.

lheilmann:
Result of checkboxes is stored in database flags.

lheilmann:
**Mockup update:**
- [ ] Team notifications
- [ ] Event notifications

birneamstiel:
@tomkellog @conradlempert I've agreed with @bakoe on how to name those flags: `team_notifications_enabled` and `event_notifications_enabled`

(see [here](https://stackoverflow.com/a/37060624) for detailed explanation)

birneamstiel:
time needed: 3.5 h

PR #456, opened 2018-01-29
Add notification settings

birneamstiel:
Is connected to: #437
issue: #301


app/views/users/notifications.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="page-header">
+  <h1><%= t 'view.user.notifications.settings' %></h1>
+</div>
+<%= form_for(@user, url: user_path(@user), method: :patch) do |f| %>
+  <div class="form-group">
+    <div class="col-lg-offset-2 col-lg-10">
+      <%= f.check_box :team_notifications_enabled %>
```
bakoe: My personal preference would be to have the `filled-in` class on these checkboxes which would make them look like this:

ohne titel 2

app/views/users/notifications.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <%= f.label :team_notifications_enabled, t('view.user.notifications.enable_team_notifications') %> <br>
+
+    </div>
+  </div>
+
+  <div class="form-group">
+    <div class="col-lg-offset-2 col-lg-10">
+      <%= f.check_box :event_notifications_enabled %>
```
bakoe: See comment to line 7

config/locales/de.yml
```diff
@@ -230,7 +230,12 @@ de:
       edit_profile: "Profilinformationen bearbeiten"
       update_profile: "Profil aktualisieren"
       favourite_sports_placeholder: "Fußball, Basketball, Tischkicker, ..."
-    
+      notifications:
+            settings: "Benachrichtigungseinstellungen"
+            enable_team_notifications: "Team Benachrichtigungen erhalten"
```
bakoe: My personal preference would be “Team-Benachrichtigungen erhalten” (with dash)

config/locales/de.yml
```diff
@@ -230,7 +230,12 @@ de:
...
+            enable_event_notifications: "Event Benachrichtigung erhalten"
```
bakoe: equivalent to comment to line 235

spec/controllers/users_controller_spec.rb
```diff
@@ -318,4 +318,20 @@
...
+      it 'should show notifications settings page successfully' do
+        sign_in @user
+        get :notifications, params: { id: @user.to_param }
+        expect(response).to be_success
+      end
+    end
+    context 'given no logged in user' do
+      it 'should deny access' do
+        get :notifications, params: { id: @user.to_param }
+        expect(response).to be_unauthorized
```
bakoe, line 332: 👍

bakoe (commented):
LGTM code-wise! Can confirm it's working in the Browser. Only some minor complaints/hints regarding design and wording

bakoe:
@birneamstiel Please make sure to integrate the latest changes from `dev` into your branch. Pay attention to the now obsolete mocked notification settings getters from `user.rb` (which **must** be removed now):
```ruby
  def has_event_notifications_enabled?
    true
  end

  def has_team_notifications_enabled?
    true
  end
```
and to the mocked setting of these settings in the tests for TeamMailer, EventMailer and MatchMailer:
```ruby
allow(user).to receive(:has_team_notifications_enabled?).and_return(true)
```
which could be replaced by something like 
```ruby
user.team_notifications_enabled = true
```
(make sure to catch all occurrences from TeamMailer, EventMailer and MatchMailer tests!)

bakoe:
ohne titel 2

@birneamstiel Also, I'm not quite sure whether „Aktualisiere Spieler“ is the right label for the button – maybe something like “Benachrichtigungseinstellungen aktualisieren” or simply “Aktualisieren” or even “Speichern” would be better?

birneamstiel:
@bakoe I've merged your changes into this branch, so maybe you could quickly checkout this branch and test your logic (if it's still working)? But I think we're good to merge :-)

bakoe (approved):
Have tested in Browser and can confirm it's working; LGTM code-wise as well!

Issue #339, opened 2018-01-14
League page: delete "Übersicht" button


svitlanakalancha:
**Übersicht** button should be deleted from the league page.

**Acceptance criteria**
- The button should be deleted **ONLY** from league page
- The list of participants should be shown on the ranking page

<img width="537" alt="sport-portal" src="https://user-images.githubusercontent.com/27727717/34919898-5806d5a6-f96a-11e7-96b0-3f288e82e55c.png">

lheilmann:
Delete route & view for overview page.

birneamstiel:
needed-time: 3h

PR #457, opened 2018-01-29
Delete overview button from event page

birneamstiel:
issue: #339


app/controllers/events_controller.rb
```diff
@@ -101,6 +101,9 @@ def ranking
   end
 
   def overview
+    if @event.type == 'League'
```
birneamstiel, line 105: @reviewers That's admittedly a bit lazy but I couldn't think of a nice way to prevent the routing only for leagues... Might also be due to the late hour, maybe there's some sudden inspiration coming over me tomorrow :-)

app/controllers/events_controller.rb
```diff
@@ -101,6 +101,9 @@ def ranking
...
+      redirect_to @event, error: 'No overview available for leagues.'
```
bakoe: Please make sure to actually use the localization you created (`overview.not_available_for_leagues`) ;-)

app/controllers/events_controller.rb
bakoe: @birneamstiel I like the approach that @jonasnoki followed in #460 – he created an explicit subclass of EventsController in order to control more precisely which requests should be routed to which controller actions (**and especially which should not**). Maybe you could take a look into his changes for inspiration, especially the following change in the routing:

```diff
-  get '/events/:id/ranking', to: 'events#ranking', as: 'event_ranking'
+  get '/leagues/:id/ranking', to: 'leagues#ranking', as: 'leagues_ranking'
```

This would make the redirection you implemented obsolete since the route would simply not exist. However, I also like your approach of redirecting the user instead of confronting him with a 404 page. Dunno … 🙃

spec/views/events/show.html.erb_spec.rb
```diff
@@ -121,9 +121,9 @@
       expect(rendered).to have_selector(:link_or_button, t('events.show.to_schedule'))
     end
 
-    it "renders an overview button" do
+    it "doesn't an overview button" do
```
bakoe: Typo: you missed "have". Should be: 

```ruby
it "doesn't have an overview button" do
```

bakoe (commented):
LGTM code-wise; haven't tested in Browser though.

app/controllers/events_controller.rb
birneamstiel: Hmm I'm not so sure about that... This seems to be a nice way of doing it but it's some serious change in structure for this small thing. I'd also have to create both `RankingListController` and `TournamentController`, correct? Or would you just move the current redirection logic into the `LeagueController`?

bakoe: Yeah, you're right – let's just stick with your implementation. 🙂

bakoe (approved):
👍  Have tested in Browser; can confirm it's working.

PR #458, opened 2018-01-29
Avoid mailer throwing delivery errors on PROD 

bakoe:
Currently, SMTP delivery errors like “550 Requested action not taken: mailbox unavailable” (caused by e. g. malformed user emails) make the deployed production websites throw errors to the user. According to https://stackoverflow.com/a/30186732, it seems common and thus reasonable to set `config.action_mailer.raise_delivery_errors` to `false` on PROD (production).


birneamstiel (approved):
LGTM! I'm no mailer expert, though :-)

Issue #218, opened 2017-12-09
Mandatory fields in team creation form


svitlanakalancha:
**Mandatory fields** in team creation form should be **recognizable**.

**Actual behaviour**

- User clicks on button "Create team"
- User sees the form
- User doesn't know what fields are mandatory to fill 
- If name isn't filled, user lands on error page

**Correct behaviour**

- User sees on the form page what fields are mandatory
- Fields are marked with * sign

niconomaa:
technical tasks

- [x] write view tests that verify presence of asterisk and footnote with explanation of mandatory fields
- [x] edit view so that it now labels obligatory fields

PR #465, opened 2018-01-29
Flexibus/#381/team page design

### LINKED ISSUE #381: Team Page Design
Neues Design von Teampage:

Contact @svitlanakalancha and team if buttons not visible (create new team as admin to see them)

Tasks:

- [x] Create team profile section.
- [x] Create kind of sport section
- [x] Create participant element
- [x] Create participant section
- [x] Create team-management section
- [x] Add buttons for participant mangement as soon as #338 has applied fixes. (Edit: we don't need to wait for #338 as that solves a different issue; we should be independent from that)
- [x] Check mobile optimizations
- [x] Fix failing tests (if any)

This includes fixing reported bug #396

- mobile version: participant view should be just one column
- 'Sportart' is just a string right now, not an array, just display the string
- 'Noch nie war es ...' should be replaced by 'description' of team
- actions should be grayed out when no participant is selected

Header and Footer are out of scope.
Buttons for group-actions are inactive (gray), if there is no player selected.

### LINKED ISSUE #396: Checkboxes on teams show page are hidden
screen shot 2018-01-18 at 17 31 50

Checkboxes are not visibile anymore after redesign was applied. 

Checkboxes are hidden, because of a missing label. Fancy checkboxes are appended as :before element to the corresponding label, that is missing. No label, no fancy checkbox!

Following code hides default checkboxes:

```
/* Remove default checkbox */
[type="checkbox"]:not(:checked),
[type="checkbox"]:checked {
  position: absolute;
  /* opacity: 0; */
  pointer-events: none;
  }
``` 

To bring back our lovely checkboxes, we  need to label them. :+1:

deeps96:
Closes #381, also fixes #396.

Make sure to checkout flexibus/**#381**/team-page-design when reviewing since there exists a branch with a similar name.


lheilmann:
I just tested locally in browser (Chrome) with the following results: 

- `/teams`: Is this the desired aligning?
screen shot 2018-01-30 at 19 28 29

- `/teams/new`: The width of the content does not fit.
screen shot 2018-01-30 at 19 36 53 1

- `/teams/1/show`: :+1:

deeps96:
@lheilmann It appears that you checked out the wrong branch. You apparently checked out https://github.com/hpi-swt2/sport-portal/pull/401 for #379. This branch is for the team**s** page (<- notice the plural), while this branch is for ticket #381, the team page design.
If you checked out the proper branch, please click on a team and review this page, only. Thanks!

The scope of this issue was only the page for a single team. It was out of scope to design the team overview page or team-creation page.

jkbe:
I'm changing the name of PR #379 to clear that up a bit at least because this has confused me before as well...

lheilmann:
@deeps96 Oh I am sorry! I misunderstood the scope of this issue/PR. I actually checked out the correct branch, but was focusing on the wrong pages. We use a different naming, e.g. teams index page, teams show page, teams edit page, ...

bakoe:
bildschirmfoto 2018-02-01 um 23 43 53


-  Should the kinds of sport (“Sportarten”) section really be displayed if the team has none assigned? (as it is the case for the team displayed in the screenshot)
- Isn't there a colon `:` missing after “… folgende Sportarten”? (compared to the mockup from the user story #465)

### Response 1

The colon is missing because of uniformity and cleaner design. @zhamanakov are you ok with that?
Is it a good idea to letting a team having no kinds of sport? I can fix this easily, but just a thought...

**ANSWER:** pls just hide the whole section, if there are no sports selected. DONE!

---

bildschirmfoto 2018-02-01 um 23 54 49

- The kinds of sport (“Sportarten”) section could need some padding on the left and right sides

### Response 1

What do you mean exactly?

### Response by @bakoe

Sorry—I have posted the wrong screenshot for the problem I described. However, the issue is rather unimportant: When there is a very long (and thus multi-line) text in the kinds of sport (“Sportarten”) section, the text has very little space (CSS: `padding`) to the left and right ends of the section.

---

bildschirmfoto 2018-02-01 um 23 47 01

- The margin between the _Select all_ and _Select none_ button seems a bit small on tablet and desktop layouts

### Response 1

@zhamanakov Is this important? If yes, how much margin do you want?

**ANSWER:** min 20px margin between buttons is important DONE

---

bildschirmfoto 2018-02-01 um 23 47 20

- The _Select none_ button is neither spaced in the same way as in the tablet/desktop layout nor right-aligned in the phone layout

### Response 1

Will fix DONE

---

bildschirmfoto 2018-02-01 um 23 54 49

- The _Select all_ and _Select none_ buttons should not be rendered when there are no checkboxes rendered 

### Response 1

Will fix that. But just for interest: When is this the case?

### Response by @bakoe

Thanks! This situation is the case whenever you look at the team page of a (public) team you are not a member of. 

DONE
---

bildschirmfoto 2018-02-01 um 23 52 05

- The font sizes of the different button groups are considerably different from each other – in the mockup from the user story #465 they all look pretty much the same

### Response 1

Will fix DONE

bakoe:
bildschirmfoto 2018-02-02 um 00 01 58

@deeps96 Is the design of the “Invite users by e-mail” in the scope of this issue as well?

bakoe:
bildschirmfoto 2018-02-02 um 00 06 44

The hover font color of the “Leave team” button has almost no contrast to the background – maybe use something lighter than black? However, white is already “reserved” for the non-hovered button text. Maybe a light blue or something like that for the hover state would be better? cc @zhamanakov 

**ANSWER:** light blue for hover will be ok! DONE

bakoe (commented):
I just skimmed the code changes since they contain a lot of HTML/CSS stuff for which – to my mind – always many “correct” solutions exist. What I looked into seemed very good. 👍 

More intensively, I tested how the re-designed page looks and behaves on Chrome and Safari on macOS. Fortunately, I couldn't find any big bugs and only noticed some minor flaws as documented in the PR's comments section.

niconomaa:
@bakoe to your q whether the email invitiation modal is part of the intended redesign: #381 does not seem to specify that explicitly, even though i'd personally find it bizarre to have this clearly unfinished element on the otherwise so nearly designed page.

### Response 1

You are right - it was out of scrope for this story. Maybe it will get an own user story during the Kanban-week? @zhamanakov 

**ANSWER:** It is ok.

---
@deeps96 even if the invitation part is out of scope for this issue, look at what happens to the top of the page after inviting a user to the team

screen shot 2018-02-02 at 08 50 20

**ANSWER:** notification re-design is already in progress in #501

jkbe:
@tomkellog #501 addresses the second issue described by you, it's out of scope for this ticket.

niconomaa:
could there be anything done about the persistent, rectangular highlighting of buttons after clicking on them?

screen shot 2018-02-02 at 08 53 40

### Response 1
I think the easiest approach would be, to remove the blue border of this selection completly. @zhamanakov are you ok with that?

**ANSWER:** yes, just remove the blue border is ok. DONE

app/assets/stylesheets/teams.css
```diff
@@ -3,12 +3,173 @@
   They will automatically be included in application.css.
 */
 
-label.required:after {
-  content: "*";
-  vertical-align: 0.5em;
-  font-size: 0.75em;
-  font-weight: normal;
```
niconomaa: the css you removed here affected the `edit` team page, adding an asterisk after required fields, as specified in #218.
please add again!
screen shot 2018-02-02 at 08 58 24

niconomaa (changes requested):

app/assets/stylesheets/teams.css
deeps96: Will fix

deeps96:
I wrote some answers into the comments. Waiting for @zhamanakov to reply - will fix everything in one go afterwards.
Also waiting for @bakoe to clarify some things.

app/assets/stylesheets/teams.css
deeps96: Done.

deeps96:
@tomkellog @bakoe @zhamanakov I fixed all things mentioned. Please recheck!

app/assets/stylesheets/teams.css
```diff
@@ -4,11 +4,192 @@
...
+    color: #1CA9E5;
+    font-weight: 500;
+}
+
+h1 {
+    margin-top: 0;
+}
+
+.genericButton, .genericButton:hover, .genericButton:focus {
+    outline: none !important;
```
bakoe: To my mind, removing the `:focus`-outline property **without adding another visual indication for focused elements** is quite an anti-pattern since it reduces the accessibility of the website significantly. See http://www.outlinenone.com for further explanation. 

Currently, the “Select all”, “Select none”, “Edit” and “Delete” buttons have a `:focus` state that looks exactly the same as the non-`:focus` state. You can inspect this e. g. by Tab-navigating (using the Tab key on you keyboard) over the Team page or by using the Chrome Developer Tools' “Force element state” panel in the Elements tab:

ohne titel

@deeps96 Please make the `:focus` states of the buttons used on the Team page distinguishable from the normal inactive state. A simple fast solution would be to extend the existing `:hover` property changes you introduced to changing not only the `:hover` state, but also the `:focus` states (e. g. by lightening up the buttons' background colors).

bakoe (approved):
@deeps96 Thanks for fixing the issues I mentioned! I took another look and can confirm that all the issues are fixed now. 

However, I have a minor complaint about the `:focus` states of some buttons being indistinguishable from the non-focused standard state (see [my comment above](https://github.com/hpi-swt2/sport-portal/pull/465#discussion_r165840651)). Maybe you can talk back to you SM/PO about the importance of accessibility on the page and decide whether you're going to fix this or not.

## Response 1
@zhamanakov Whats your opinion?
I would be happy to merge this, as soon as @tomkellog approves that the requested changes from his review has been implemeneted.

niconomaa (approved):

jkbe:
I think @zhamanakov needs to decide on the `:focus` issue as it's a matter of accessibility and therefore product requirements. Personally, I can understand both the notion of a clean design as well as that of accessibility and supporting different methods of navigating a webpage. Unfortunately, the [PO notes on non-functional requirements](https://github.com/hpi-swt2/sport-portal/wiki/Non-Functional-Requirements) don't give any input on this either.

bakoe:
@jak-ing @deeps96 @zhamanakov 

Have you thought about the issue I mentioned in the user story?

https://github.com/hpi-swt2/sport-portal/issues/381#issuecomment-362432106

> While reviewing the PR, I noticed that there is no visual indication anymore about whether a team member is a team owner or not. Before the redesign, the user could somehow guess this based on the enabled/disabled state of the “Add ownership” and “Remove ownership” buttons – at least, if the user was the team owner.
> 
> Is it intended that there is absolutely no visual indication anymore showing which team member is a team owner?

## Response 1
Still waiting for @zhamanakov response.

Issue #462, opened 2018-01-29
ApplicationMailer is not able to send e-mails via the GMX account provided


niconomaa:
Currently, the error `550-Requested action not taken -- mailbox unavailable` is thrown when automated email notifications are sent. 
Further configuration is needed to be able send mails.

bakoe:
It turned out that `ApplicationMailer`'s `default from: 'from@example.com'` had overwritten the ENV variable `SENDER_EMAIL` (sport.portal@gmx.de) which made GMX return a 550 error. Furthermore, the error was logged in Rails only with the quite unspecific “mailbox unavailable” message; when I tried to send an email via SMTP manually in a SMTP client CLI, I also got a much more specific “Sender address is not allowed” message which pointed into the right direction to fix the bug.

bakoe:
Sadly, the mails do again not work at the moment (Jan 30th 3 p.m.) This may need some more work …

bakoe:
… aaaaand it works again. GMX seems to have had some problems.

bakoe:
Actual time I invested in this issue: 1 h

PR #466, opened 2018-01-29
Make ApplicationMailer able to send e-mails via the configured SMTP server

bakoe:
See #462

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,6 +1,5 @@
 # An ActionMailer sublcass that is used for all email notifications triggered on the platform
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
   layout 'mailer'
 
   #General mailer class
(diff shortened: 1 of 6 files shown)


jkbe (approved):
Please, split up these huge PRs in the future!

nikriek:
Bye bye coverage

nikriek:
Is that enough to fix the issue?

jkbe:
Whoops, I approved just before the Travis Builds failed. Seems the Application Mailer needs some default so that the sender address is not `nil`. Just set it to the GMX Sport Portal address as well?

bakoe:
@nikriek 
> Is that enough? 

It seems like it's enough – at least locally the e-mails can be sent via the GMX account now. 🙂

bakoe:
@jak-ing 
> Just set it to the GMX Sport Portal address as well?

I'm on it. 👍

birneamstiel (approved):

PR #491, opened 2018-01-31
Confirm a game result (Ref #236)

nikriek:
Connects #236


app/models/event.rb
```diff
@@ -21,12 +21,6 @@
 #  selection_type       :integer          default("fcfs"), not null
 #  min_players_per_team :integer
 #  max_players_per_team :integer
-#  matchtype            :integer
```
citron0xa9: why is this deleted?

app/models/game_result.rb
```diff
@@ -2,16 +2,31 @@
...
+    !(is_confirmed? || match.users_in_same_team([user, scores_proposed_by]))
+  end
+
+  def is_confirmed?
+    scores_proposed_by.blank?
+  end
+
+  def confirm_scores
+    self.scores_proposed_by = nil
+    save
```
citron0xa9: is the save here really necesarry? / I think it is not the responsibility of the method to save the model.

app/models/league.rb
```diff
@@ -21,12 +21,6 @@
 #  selection_type       :integer          default("fcfs"), not null
 #  min_players_per_team :integer
 #  max_players_per_team :integer
-#  matchtype            :integer
```
citron0xa9: same as above

app/views/matches/_edit_results.html.erb
```diff
@@ -22,12 +22,19 @@
               <%= builder.number_field :score_away, in: -2e9..2e9, step: 1, class: 'form-control' %>
             </div>
           </div>
+          <% unless builder.object.new_record? %>
+            <div class="form-group">
+              <% if can? :confirm, builder.object %>
+                <%= link_to t("view.match.confirm_game_result"), confirm_game_result_match_path(id: @match.id, result_id: builder.object.id), class: 'btn btn-default' %>
```
citron0xa9: I would add `col-xs-offset-1` so the layout looks better

citron0xa9 (changes requested):
Some general comments:
 - looks good mostly
 - the `col-xs-offset-1` thing is the only change I really request, I am ok with the rest / it is probably PO related review
 - a unconfirmed result already effects schedule, point calculation etc. I don't know if this is wanted
 - there is no feedback when clicking the confirm button
 - somewhat related to this issue: for rankinglist i did not find a way to view matches and therefore no way to confirm results

app/models/event.rb
nikriek: I just executed `$ annotate`, will do again to make sure it's up to date

app/models/game_result.rb
nikriek: Should be in the controller, good point 👍

app/views/matches/_edit_results.html.erb
nikriek: Cool :)

features/step_definitions/meta.rb
```diff
@@ -37,6 +37,7 @@
 end
 
 Then(/^there should be a back button on all relevant pages$/) do
+  puts page
```
manner: Should this be here?

spec/controllers/matches_controller_spec.rb
```diff
@@ -288,6 +281,7 @@
         result = match.game_results.first
         expect(result.score_home).to eq(new_attributes["0"]["score_home"])
         expect(result.score_away).to eq(new_attributes["0"]["score_away"])
+        expect(match.scores_proposed_by).to eq(@user.teams.where('team_id IN (?)', match.teams).first)
```
manner: you can use `where(id: match.teams).first`

app/models/match.rb
```diff
@@ -230,4 +232,20 @@ def get_match_elo_result(home, away)
   def teams
     [team_home, team_away]
   end
+
+  def propose_scores(user)
+    self.scores_proposed_by = user.teams.where('team_id IN (?)', self.teams).first
```
manner: you can use `where(id: self.teams).first`

app/models/match.rb
```diff
@@ -230,4 +232,20 @@ def get_match_elo_result(home, away)
...
+  end
+
+  def can_confirm_scores?(user)
+    !(scores_confirmed? || user.teams.where('team_id IN (?)', self.teams).first == self.scores_proposed_by)
```
manner: same here: you can use `where(id: self.teams).first`

app/models/match.rb
nikriek: Good point!

manner:
When a team is supposed to confirm a match and accidentally presses the SAVE button instead of the CONFIRM button, then it can no longer confirm the match. Then neither of the teams can confirm a match anymore

citron0xa9 (approved):
Please check comments of @manner before merge

Issue #332, opened 2018-01-14
Players should only be able to edit the matches in which they played


PoJo93:
**AS** player of a match, organizer of an event(or admin) and no one else
**I WANT TO** be able to edit the scores of the matches in which I played in
**TO** ensure that no foreign users edit them 

**ACCEPTANCE CRITERIA**
- [x] When I am not an eligible person( see top) I don't have the option to edit or enter scores for a match
   - [x] When I see the playing schedule of a event I don't see a button to submit a new score nor do I see the textboxes to enter values
  - [x] When I go in the settings of a match I am just able to see the scores entered for a team but not able to suggest new ones or edit the current ones
  - [x] I can just see the entered scores and their status (confirmed by both teams or just suggested by one team)
- [x] When I am a eligible person I have the already provided functionality of editing scores

PR #492, opened 2018-01-31
WIP - Players can only edit own matches (Ref #332)

manner:


app/controllers/matches_controller.rb
```diff
@@ -48,10 +49,12 @@ def update_points
 
   # GET /matches/1/edit_results
   def edit_results
+    authorize! :edit, @match
   end
 
   # PATCH/PUT /matches/1/update_results
   def update_results
+    authorize! :edit, @match
```
nikriek: You can use authorize as a global controller function

nikriek (changes requested):

app/controllers/matches_controller.rb
```diff
@@ -1,5 +1,6 @@
 class MatchesController < ApplicationController
   before_action :set_match, only: [:show, :edit, :update, :update_points, :edit_results, :update_results, :destroy, :add_game_result, :remove_game_result]
+  authorize_resource only: [:edit]
```
nikriek, line 3: Why only edit?

manner: US 332 only specifies to restrict editing

nikriek: then create a ticket? maybe ask @PoJo93

spec/controllers/matches_controller_spec.rb
```diff
@@ -33,13 +53,53 @@
...
+    it "shouldn't allow other users to edit the results of a match" do
+      sign_out @user
+      sign_in @other_user
+      match = Match.create! valid_attributes
+      get :edit_results, params: { id: match.to_param }
+      expect(response).not_to be_success
+    end
+
+    it "should allow admin to edit the results of a match" do
+      sign_out @user
```
nikriek: Do you really need to sign out the other user?

nikriek (approved):

Issue #308, opened 2018-01-10
Notifications: events/matches


svitlanakalancha:
As User I want to be informed about changes in my events to plan my matches accordingly.

**Acceptance criteria**
- general requirements see #301 
- emails are automatically generated when triggered
- the text of emails is predefined and includes placeholders for customization (see texts below)

Triggers are:
- a match where user participates is scheduled

> Hallo _user name_,
> dein Spiel gegen _oponent team_ in _event name_ findet am _datum_ statt.
> Viel Erfolg beim Training!
> _Link to the event_

- a match where user participates is cancelled or postponed

> Hallo _user name_,
> dein Spiel gegen _oponent team_ in _event name_ wurde (auf  _datum_ verschoben / abgesagt).
> Viel Erfolg beim Training!
> _Link to the event_

- an event where user is registered is started/cancelled/finished

> Hallo _user name_,
> das Event _event name_ hat gestartet (ist abgesagt/ ist abgeschlossen).
> Viel Erfolg beim Training!
> _Link to the event_

**Important**
Check issues #57 and #149 and communicate with other dev teams

niconomaa:
### triggering actions based on model events
* [this page](http://guides.rubyonrails.org/active_record_callbacks.html) describes Active Record Callbacks, which are methods that can be triggered on several model events such as delete, create and update
* for events such as de deletion or creation of a match a user participated in, these callbacks can be used to notify the user
* in the model at hand a callback may be implemented like this:
```
  after_create :notify_about_member
 
  def notify_about_member
    puts 'Hey ' + User.find(self.user_id).first_name + ", you are part of the team now!"
  end
```
### triggering actions at certain times
* [this page](http://edgeguides.rubyonrails.org/active_job_basics.html) describes Active Jobs, which are a class in Rails that define jobs to be executed at a certain time
* combining Active Record Callbacks with Active Jobs, we could, for instance, schedule jobs when each match is created for the time in happens, notifying the members of each team that participats int the match about it
* a job that notifies a given member about his or her match might be implemented like so:
```
class NotifyUserJob < ApplicationJob
  queue_as :default

  def perform(user)
    puts User.find(user).first_name + " your match is NOW!"
  end
end
```
* and a method in a model that schedules the job might look like this:
```
    # SPIKE
    after_create :queue_member_notification
 
    def queue_member_notification
      NotifyUserJob.set(wait: 15.seconds).perform_later(User.first.id)
    end
```

[...]

Issue #307, opened 2018-01-10
Notifications: team


svitlanakalancha:
As User I want to be informed about changes in my teams to be better prepared for coming sports events.

**Acceptance criteria**

- general requirements see in the #301 
- emails are automatically generated when triggered
- the text of emails is predefined and includes placeholders for customization (see texts below)

Triggers are:

- User is added to a team and receives correspondent email.

> Hallo _user name_,
> Du bist dem Team _team name_ beigetreten.
> ODER Deine Teamliste wurde aktualisiert.
> Viel Erfolg beim Sport!
> Platform name team

- Team where user participates was registered to some event

> Hallo _user name_,
> Dein Team _team name_ nimmt jetzt an dem Event _link_ teil.
> Viel Erfolg beim Spielen!
> Platform name team

**Important**
Check issues #57 and #149 and communicate with other dev teams

bakoe:
## Clarification needed

- [X] @svitlanakalancha: Should the “Team where user participates was registered to some event” notification be controlled by the user's “Team notifications” or by his/her “Event notifications”? 
  - --> stays at “Team notifications”

## Development Tasks

- [x] Create a `TeamMailer` inheriting from `ApplicationMailer` via the corresponding rails CLI command
- [x] Add `user_added_to_team` method to `TeamMailer` that prepares emails following the template from the corresponding AC
  - Bonus: Send the mail not to `hans@example.org` but to `Hans Mustermann <hans@example.org>` – see http://guides.rubyonrails.org/action_mailer_basics.html#sending-email-with-name
- [x] Add private `prevent_delivery_to_unsubscribed_users` method to `TeamMailer` model that checks the user's `team_notifications_enabled` (see #301) flag and disables the delivery when flag is set to false – see http://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks
- [x] Add `after_action` hook to `TeamMailer` model which triggers the `prevent_delivery_to_unsubscribed_users` method
- [x] Add `team_registered_to_event` method to `TeamMailer` model that prepares emails following the template from the corresponding AC
  - Make sure that the link URL is generated correctly – see http://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views
  - Bonus: Send the mail not to `hans@example.org` but to `Hans Mustermann <hans@example.org>` – see http://guides.rubyonrails.org/action_mailer_basics.html#sending-email-with-name
[...]

PR #496, opened 2018-01-31
custom have fun messages for team mails. ref #308

conradlempert:


birneamstiel (approved):

app/views/team_mailer/team_registered_to_event.html.erb
```diff
@@ -2,5 +2,5 @@
 <p><%= t('.registered_to_event_info_html',
          team_link: link_to(@team.name, team_url(@team)),
          event_link: link_to(@event.name, event_url(@event))) %></p>
-<p><%= t('mailer.have_fun') %></p>
+<p><%= t('.have_fun') %></p>
```
bakoe, line 5: Nice find that we can use `.have_fun` without explicitly defining e. g. `user_added_to_team` as the locale to look into for the localization. 👍

bakoe (approved):
LGTM! 👍 I have also confirmed that the change is working by using `rails c` to provoke both e-mail notification events – the localizations are used as one would expect.

bakoe (changes requested):
Although not explicitly defined in the initial user story #308 or in @svitlanakalancha's change request stated in Slack (see channel #307-308-coordination), I'm pretty sure that @svitlanakalancha especially wanted that User's don't get the ”Viel Erfolg beim Training!“ message when matches are canceled and when events are canceled/finished. Could you maybe integrate this in this PR as well @conradlempert?

bakoe:
@conradlempert See the following message from the Slack channel for explanation about what I meant regarding @svitlanakalancha's change request:

ohne titel 3

Maybe just remove the `have_fun` part from the notifications about the following events?

- match being canceled
- event being canceled
- event being finished

bakoe (approved):

PR #502, opened 2018-01-31
Swteam/#375/working back buttons

### LINKED ISSUE #375: Working "back"-buttons
**As a user** of the platform
**I want to** have intuitively working "back"-buttons
**In order to** navigate the platform easily.

## Acceptance Criteria:
 * It should be possible to get back from every subpage to its parent page without detours
    * example: event schedules and overview pages have no back button
 * Existing "back"-Buttons should refer to the expected page not the last visited page
    * example: Try editing an event and then clicking on back, currently you come back to the editing page but you should get back to the all events page
 * If you are not sure what the right page for referring to is, ask your PO :D

### Notes:
 * mentioned as bug #158 
 * It's probably wise to find a solution that works for all pages and does not require you to enter this data on every single page manually

Oluwoye:
There were major issues with Travis. Still needing feedback from PO.

Will close #375


Gemfile
```diff
@@ -170,7 +170,7 @@ end
 group :production do
   # Ruby interface to the PostgreSQL RDBMS
   # https://github.com/ged/ruby-pg
-  gem 'pg'
+  gem 'pg', '~> 0.18'
```
fpottbaecker: Why this change in version?

Gemfile.lock
```diff
@@ -214,7 +214,6 @@ GEM
     parallel (1.12.1)
     parser (2.4.0.2)
       ast (~> 2.3)
-    pg (0.21.0)
```
fpottbaecker: Why is this no longer present?

db/schema.rb
```diff
@@ -82,13 +82,6 @@
     t.index ["user_id"], name: "index_organizers_on_user_id"
   end
 
-  create_table "participants", force: :cascade do |t|
```
fpottbaecker: Why is this table gone?

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
...
+    # There are a lot of routs in Rails, that can't be visited in general
+    begin
+      visit currentpath
+    rescue Exception
+      currentpath = '/imprint'
+      visit '/imprint'
+    end
+    puts currentpath
+    unless (excludedrouts.include? currentpath) || (temporarywedontknowwhattodorouts.include? currentpath)
+       expect(page).to have_link('Zurück')
```
fpottbaecker: I18n.t please

features/back_buttons/present_buttons.feature
```diff
@@ -0,0 +1,12 @@
...
+  As a user of the platform
+  I want to have working "back" buttons
+  In order to navigate through the platform easily
+
+  Scenario: Visit all pages
+    Given a user
+    And the user is logged in
+    And a tournament 'MyTournamentus' with 4 teams
+    And a tournament 'Tourney2' with 2 users
+    And the Spielplan page for 'MyTournamentus' is visited
```
fpottbaecker: This scenario looks a little bloated, why do we need to visit the tournaments page?

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
   }
   expect(page).to have_xpath("//h1[contains(text(),'#{I18n.t *translations[text]}')]")
 end
+
+
+Then(/^there should be a back button on all pages except the start page$/) do
```
fpottbaecker: I think this should be an RSpec test so you can have a list of all relevant routes in your test ui.

fpottbaecker (changes requested):

Gemfile
Oluwoye: Because Travis load the 1.0.0 version automatically and failed with it. After I disabled pg at all, Travis said, that he wants to have version 0.18

db/schema.rb
Oluwoye: Fixed

features/step_definitions/content.rb
Oluwoye: We thought about this, but this is all functionality, that we might test: What do you think is best: Only a Cucumber test, only a RSpec test or a Cucumber and a redundant RSpec test?

features/back_buttons/present_buttons.feature
Oluwoye: To create at least one match and a complete schedule.

Gemfile.lock
Oluwoye: Fixed

features/step_definitions/content.rb
Oluwoye: Applied

app/views/users/edit_profile.html.erb
```diff
@@ -0,0 +1,59 @@
+<%= render partial: "application/validation_errors", locals: {model: @user} %>
```
ErikBrendel: why is there a whole new file?

app/views/events/ranking.html.erb
```diff
@@ -34,3 +34,7 @@
     <% end %>
   </tbody>
 </table>
+
+<%= link_to t('helpers.links.back'),
```
ErikBrendel: these calls are duplicating a bit. You could maybe extract this functionality to `ApplicationHelper::back_button_to(url)`

Would in this case look like `<%= back_button_to event_path(@event) %>`

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
...
+  Rails.application.routes.routes.each do |route|
+    currentpath = route.path.spec.to_s
+    if currentpath.include? '(.:format)'
+      currentpath['(.:format)'] = ''
+    end
+    if currentpath.include? ':id'
+      currentpath[':id'] = '1'
+    end
+
+    # There are a lot of routs in Rails, that can't be visited in general
```
ErikBrendel: *`routes`

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
...
+    begin
+      visit currentpath
+    rescue Exception
+      currentpath = '/imprint'
```
ErikBrendel: what?

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
...
+      visit '/imprint'
+    end
+    puts currentpath
```
ErikBrendel: debug output should be removed

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
...
+    unless (excludedrouts.include? currentpath) || (temporarywedontknowwhattodorouts.include? currentpath)
+       expect(page).to have_link(I18n.t ('helpers.links.back'))
+    end
+  end;nil
```
ErikBrendel: wow, what is that?

ErikBrendel (changes requested):

app/views/users/edit_profile.html.erb
Markus-Brand: I think this file was deleted by us (merged with another file) and merge conflicts somehow recovered it.
You can just delete it, i guess

features/step_definitions/content.rb
```diff
@@ -55,3 +55,33 @@
   }
   expect(page).to have_xpath("//h1[contains(text(),'#{I18n.t *translations[text]}')]")
 end
+
+
+Then(/^there should be a back button on all pages except the start page$/) do
+  # These routes were excluded in accordance with our PO
+  excludedrouts = ['/', '/cable', '/tournaments/new', '/leagues/new', '/leagues/1/edit', '/tournaments/1/edit', '/events/1/edit', '/rankinglists/new', '/teams/new', '/teams/1/edit', '/matches/1/edit', '/my/users/sign_in', '/my/users/password/edit', '/my/users/cancel', '/my/users/sign_up', '/users/1/dashboard', '/users/1/edit', '/matches/1/add_game_result', '/matches/1/edit_results']
```
Markus-Brand: multiple cases of "routs" instead of "routes" here. Also please rename your variables to be seperated by underscores like this: excluded_routes

features/step_definitions/content.rb
Oluwoye: Can you explain your problem a bit more?

features/step_definitions/content.rb
Oluwoye: The test searches all routes accessible in Rails. Not all of these routes are visitable with "visit page". But at all pages, which weren't excluded by PO and are visitable should be a back button.

features/step_definitions/content.rb
Oluwoye: Fixed

features/step_definitions/content.rb
Oluwoye: Yep done

features/step_definitions/content.rb
Oluwoye: Done

app/views/users/edit_profile.html.erb
Oluwoye: Me or Erik?^^ And should I remove this linke or the whole file? Where is  the new file for this page?

features/step_definitions/content.rb
ErikBrendel: why are you redirecting to the imprint in the error case? That does not make any sense to me

features/step_definitions/content.rb
ErikBrendel: I mean the `end;nil` part :D

features/step_definitions/content.rb
Oluwoye: Ohh good point! Fixed!

features/step_definitions/content.rb
Oluwoye: Ohh ... yep^^

Oluwoye: Fixed ^^

features/step_definitions/content.rb
fpottbaecker: Ideally, just an rspec test, this seems to granular for an rspec test.

app/views/users/edit.html.erb
```diff
@@ -120,6 +120,7 @@
       <%= link_to t('.cancel', default: t('helpers.links.cancel')),
                 user_path(@user), class: 'btn btn-default' %>
       <%= t('.unhappy') %>? <%= link_to t('.cancel_this_account'), user_path(@user), method: :delete %>.
+      <%= back_btn root_path %>
```
ErikBrendel: @Oluwoye I deleted `users/edit_profile`, since this file got merged into `users/edit` (as you see here). Is this one linge the only change you made to `users/edit_profile?`

ErikBrendel (approved):

app/views/users/edit.html.erb
Oluwoye: yes

features/step_definitions/content.rb
Oluwoye: Do you mean "this seems too granular for a cucumber test"? Then I would copypaste all lines out of the step definitions, create a rspec with it and delete this test?

fpottbaecker (approved):

DerIntergalaktischeGrosshorst (approved):

Issue #489, opened 2018-01-31
Saving notification settings lands on dashboard


svitlanakalancha:
**Actual behavior**
- Sign in and open `Benachrichtigungseinstellungen`
- Change some settings and click `Aktualisieren`
- Dashboard page is shown

**Correct behavior**
- `Benachrichtigungseinstellungen` page is shown after updating settings

PR #503, opened 2018-02-01
Stay on Notification Settings page after applying settings

birneamstiel:
see #489

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -334,4 +334,13 @@ RSpec.describe UsersController, type: :controller do
     end
   end
 
+  describe 'PATCH #update_notifications' do
+    it 'should update notifications settings successfully' do
+      sign_in @user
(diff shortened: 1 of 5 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -334,4 +334,13 @@
     end
   end
 
+  describe 'PATCH #update_notifications' do
+    it 'should update notifications settings successfully' do
```
bakoe, line 338: Maybe consider adding a separate controller/view/feature test that checks whether the user is actually redirected to the notification settings page after he clicked the “Save” button on the page. Otherwise, issue #489 would technically not be represented by a test case. :wink:

bakoe (commented):
LGTM code-wise and can confirm it's working in the Browser. 👍 

However, I'd like to have an additional test that explicitly checks the expected behavior stated in #489 (see review comment).

bakoe (approved):
Cool, thanks for adding the test! 👍

jonasnoki (approved):
This should close the issue! LGTM 👍

Issue #488, opened 2018-01-31
Single Players should not be shown in Teams


svitlanakalancha:
This bug is caused by the other bug, namely: if a single user joins some event, the team with only this user is automatically created. This "team" will be shown then on the teams' list.

**Correct behavior**
Single player teams should **not be shown** on the `Teams` page.

birneamstiel:
only on index page, no change on email logic

niconomaa:
#assi

PR #507, opened 2018-02-01
Hide single-player "fake" teams on Teams index page

bakoe:
See #488

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,7 +78,7 @@ class User < ApplicationRecord
 
   has_many :team_users
   has_many :teams, through: :team_users, source: :team
-  has_many :real_teams, -> { where single: false }, through: :team_users, source: :team
+  has_many :non_single_teams, -> { where single: false }, through: :team_users, source: :team
   has_many :team_owners, -> { where is_owner: true }, source: :team_user, class_name: "TeamUser"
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -78,6 +78,7 @@ def four_different_characters
 
   has_many :team_users
   has_many :teams, through: :team_users, source: :team
+  has_many :real_teams, -> { where single: false }, through: :team_users, source: :team
```
niconomaa: maybe naming could be more descriptive than "real" as in "non-single-teams" or such...

niconomaa (approved):

app/models/user.rb
bakoe: Good idea—I was not very happy with the name either. I'm gonna change it right now. 🙂

bakoe: Resolved in 329929b

app/models/user.rb
```diff
@@ -78,6 +78,7 @@ def four_different_characters
 
   has_many :team_users
   has_many :teams, through: :team_users, source: :team
+  has_many :non_single_teams, -> { where single: false }, through: :team_users, source: :team
```
jonasnoki, line 81: Good idea!!

jonasnoki (approved):
I like your code 💻

PR #508, opened 2018-02-02
Kickerz/hide members for not logged in

niconomaa:
Hide form containing table of team members on team show page from users who are not logged in.
@reviewers: Use https://github.com/hpi-swt2/sport-portal/pull/508/files?w=1 to see the actual diff without whitespace changes.


spec/features/team/show_team_spec.rb
```diff
@@ -61,4 +61,19 @@
       expect(page).to have_link(I18n.t("helpers.links.email"), href: mailto_link_for_user)
     end
   end
+
+  describe 'display team members only for logged in users' do
+
+    it 'should not display team member table for users who are not logged in ' do
+      sign_out @user
```
lheilmann: @tomkellog I would assume `@user` is not signed in, hence there is no need to sign out the `@user`. Can this line be deleted?

lheilmann: Or you add the following block and adjust both test cases:
```
before(:each) do
    sign_in @user
end
```

spec/features/team/show_team_spec.rb
```diff
@@ -61,4 +61,19 @@
...
+      visit team_path @team
+      expect(page).to_not have_css("table.table-striped")
```
lheilmann: @tomkellog May not be the best approach to check on a table with no specific id. Imagine there are going to be multiple tables with the css-class `.table-striped` on this page... What about giving the table a meaningful id?

lheilmann (commented):
Some minor thoughts. Rest looks good to me :beach_umbrella: Tested in browser, too!

bakoe (commented):
LGTM code-wise, haven't tested in Browser though – I agree with the changes requested by @lheilmann.

bakoe (approved):

spec/features/team/show_team_spec.rb
```diff
@@ -61,4 +61,18 @@
...
+
+    it 'should not display team member table for users who are not logged in ' do
+      visit team_path @team
+      expect(page).to_not have_css("table#members")
+    end
+
+    it 'should display team members table for users who are logged in' do
+      sign_in @user
+      visit team_path @team
+      expect(page).to have_css("table.table-striped")
```
bakoe: Please use the newly introduced CSS ID here as well.

bakoe (changes requested):

lheilmann (commented):
You will have my GO, when you applied the changes @bakoe has requested.

spec/features/team/show_team_spec.rb
niconomaa: ooooops

niconomaa: done

lheilmann (approved):
1A!

bakoe (approved):

Issue #317, opened 2018-01-13
No leaving event after deadline


GittiHab:
**AS AN** organiser of an event
**I WANT TO** not allow participants to leave events after the deadline
**SO THAT** the schedule is not messed up

## ACCEPTANCE CRITERIA
- [x] After the deadline of an event no one can leave events anymore 
- [x] The leave button is hidden for all users

GittiHab:
Looks good!

PR #516, opened 2018-02-02
Make events only leavable if their deadline has not passed.

stlutz:
Does not apply to ranking lists.

Ref: 
#317


Jeilef (approved):
git gud

PR #518, opened 2018-02-02
Issuenumber5/#136/elo ranking

### LINKED ISSUE #136: Elo Ranking
**AS AN** organiser of a ranking list event
**I WANT TO** be able to select elo ranking as a metric
**SO THAT** the users can be ranked by this metric

## ACCEPTANCE CRITERIA
- [x] When creating a new event of the type "Ranking List" there should be a metric option "Elo"
- [x] When a user added a new match to a ranking list event the system should calculate the new elo score of the participants based on the metric of the event
- [x] The new score is calculated based on the [elo number](https://en.wikipedia.org/wiki/Elo_rating_system)
- [x] The new score of each participant is saved so that it can be displayed in the event evaluation

pawelsoundso:
Apply PO-Feedback. Closes #136.


Jeilef (approved):
gut

PR #520, opened 2018-02-03
Flexibus/#500/edit profile ui changes

### LINKED ISSUE #500: Small improvements in "User profile edit"
- [x]  Input-text should have the same size as default text.
- [x]  Label and text-input should be in one line. 

- [x]  Telegram Logo should be deleted:

- [x]  Small changes in buttons from below of the page:

jkbe:
Closes #500.
- buttons to save changes on edit pages now show "Speichern" instead of "Aktualisiere \<model\>" everywhere
- label and text-input sizes and alignment should be fixed for all forms, not just the one on the edit user page

### DIFF
--- a/config/locales/de.helpers.yml
+++ b/config/locales/de.helpers.yml
@@ -40,7 +40,7 @@ de:
     submit:
       create: "Erstelle %{model}"
       submit: "Speichere %{model}"
-      update: "Aktualisiere %{model}"
+      update: "Speichern"
       delete: "Lösche %{model}"
     select:
       prompt: "Bitte wähle"
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -70,7 +70,7 @@ body {
     padding-bottom: 120px;
   }
 
-  * {
+  *, .form-horizontal .form-group input.form-control {
     font-size: 1.2rem;
   }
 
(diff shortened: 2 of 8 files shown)


config/locales/de.helpers.yml
```diff
@@ -40,7 +40,7 @@ de:
     submit:
       create: "Erstelle %{model}"
       submit: "Speichere %{model}"
-      update: "Aktualisiere %{model}"
+      update: "Speichern"
```
jonashering, line 43: Creates inconsistency between `en.helpers.yml` and `de.helpers.yml`; One requires a model to be set, and one doesn't. Although current focus is on the German version, I guess it's still a good idea to keep existing files clean and up to date. 

Why don't you use the already existent text from `t'(helpers.links.save)` or change the line in all concerned locale files.

jonashering (changes requested):
The redesign looks neat! Great work, Jakob! 💯 
Just have a small comment...

config/locales/de.helpers.yml
jkbe: Yup, I'll adapt the `de.helpers.yml` accordingly.
I decided against using `helpers.links.save` because then I'd have to pass this string manually on all edit pages for all models vs just changing the string which is automatically retrieved in Rail's convention-over-configuration manner when it sees an `<%= f.submit ... %>` on an edit page.

jkbe:
Please recheck @jonashering

config/locales/de.helpers.yml
jonashering: Okay, I agree.

jonashering (approved):

PR #522, opened 2018-02-03
Swteam/#486/general contact info

### LINKED ISSUE #486: add option to add general contacting info to a profile
**As an** user
**I want to** have an additional contact field on my profile page
**In order to** have the possibility to add general contacting information

### notes
this was part of #369 

## Acceptance criteria:
 * [ ] please add an additional text field for general contacting possibilities, e.g. (1) (would be nice if the fields content would be visible on the profile page as well =) )

### pictures:
(1)
profileediting

ErikBrendel:
With this PR, Users have a text field for storing additional contact information. It gets displayed on their profile if not empty (empty by default).

There are no validations around that String (length / special characters). 

-----

closes #486


app/views/devise/registrations/show.html.erb
```diff
@@ -25,6 +25,10 @@
     <dt style="width:165px"><strong><%= model_class.human_attribute_name(:favourite_sports) %>:</strong></dt>
     <dd><%= @user.favourite_sports %></dd>
   <% end %>
+  <% if @user.contact_information.present? %>
+      <dt style="width:165px"><strong><%= model_class.human_attribute_name(:contact_information) %>:</strong></dt>
```
MerlindlH: the indentation differs from the rest...

MerlindlH (commented):
looks good to me

app/views/devise/registrations/show.html.erb
ErikBrendel: Yes :)
But what are you doing here?

ErikBrendel: Wobei dich womöglich der Inhalt der yaml Dateien interessiert :D

spec/models/user_spec.rb
```diff
@@ -238,4 +238,13 @@
     email_with_name = user.email_with_name
     expect(email_with_name).to eq("'Hans Mueller' <hans@example.org>")
   end
+
+  it 'can keep general contacting information' do
```
fpottbaecker, line 243: This basically tests that rails can store attributes. Couldn't you just test, whether it is valid when the contact information is empty?

spec/views/users/show.html.erb_spec.rb
```diff
@@ -1,16 +1,29 @@
...
+  it 'displays contact information if available' do
+    contact_information = 'My homepage: foo-bar.com (check it out! :D)'
+    @user.contact_information = contact_information
+    render
+    expect(rendered).to have_content(I18n.t('helpers.label.user.contact_information'), count: 1)
+    expect(rendered).to have_content(contact_information, count: 1)
+  end
+
+  it 'does not show any contact information if none is available' do
+    @user.contact_information = nil
```
fpottbaecker, line 26: `render`?

config/locales/de.helpers.yml
```diff
@@ -71,6 +71,7 @@ de:
         telephone_number: "Telefonnummer"
         telegram_username: "Telegram Benutzername"
         favourite_sports: "Lieblingssportarten"
+        contact_information: "Kontaktinformationen"
```
fpottbaecker: Duplicate translation, and unused as well.

fpottbaecker (changes requested):

spec/models/user_spec.rb
ErikBrendel: It failed without the migration. I think this is a valid model test. But yes, I could also test if the user is valid.

spec/views/users/show.html.erb_spec.rb
ErikBrendel: Oh yes, I forgot that. Will add it of course.

ErikBrendel: done

DanielStephan97 (approved):

app/views/devise/registrations/show.html.erb
```diff
@@ -34,35 +32,18 @@
         <p class="description-text"><%= @user.favourite_sports %></p>
     <% end %>
   </div>
+  <div class="row">
```
fpottbaecker, line 35: Shouldn't the if be outside of the `row`? same on all of the above, so i would ask the design team @hpi-swt2/team-dz-17

spec/views/users/show.html.erb_spec.rb
```diff
@@ -18,12 +18,13 @@
...
-    expect(rendered).to have_content(I18n.t('helpers.label.user.contact_information'), count: 1)
+    expect(rendered).to have_content(I18n.t('activerecord.attributes.user.contact_information'), count: 1)
     expect(rendered).to have_content(contact_information, count: 1)
   end
 
   it 'does not show any contact information if none is available' do
     @user.contact_information = nil
-    expect(rendered).to have_content(I18n.t('helpers.label.user.contact_information'), count: 0)
+    render
+    expect(rendered).to have_content(I18n.t('activerecord.attributes.user.contact_information'), count: 0)
```
fpottbaecker: I would suggest `to_not have_content(User.human_attribute_name(:contact_information))`

app/views/devise/registrations/show.html.erb
ErikBrendel: have you contacted any of them directly? You know, many of the developers have dis-watched this repo because of the notification mess

jkbe: Yeah, I think it's better to place the `<%if ... end %>` outside of the row so that no empty row is rendered. Don't you end up with a bunch of whitespace this way? (Haven't checked out the branch myself so far.)

fpottbaecker: This is not a problem of this branch, all attributes are wrapped like this.

fpottbaecker (approved):

Issue #494, opened 2018-01-31
Single player team should not receive team notifications


svitlanakalancha:
This bug is caused by the other bug, namely: if a single user joins some event, the team with only this user is automatically created. This "team" should not receive team notifications.

lheilmann:
**Development tasks:**
- [x] Add tests for team notifications in `spec/models/team_spec.rb` regarding single player teams
- [x] Implement checks before delivering notification mails to prevent single player teams to receive mails

lheilmann:
@svitlanakalancha To summarize: Single player teams should not receive team notifications at all, but teams with 2 or more members should be able to receive team notifications as discussed in #307?

svitlanakalancha:
@lheilmann The issue is not really about the number of players in the team. These teams created for single players who joined some event are not real teams: you can´t add other users there. It is kind of stable team with one player. These fake teams should not receive emails. But the team created in a normal way as a team should receive emails even if it consists of one player only. Hope it helps :)

lheilmann:
@svitlanakalancha Alright! Thanks for clarification :+1: 

@pawelsoundso Hi Pawel! You were implementing the `single` column on teams a month ago right? I assume the intention was to allow single players to attend events as well, without changing the already built infrastructure completely. Hence you needed to create teams flagged with a boolean to distinguish between multi-player teams and single-player teams? Please tell me, if I understood something wrong. 

My question to you regarding this issue is the following: Are there any other usages of this flag `single` or is it only needed in your events infrastructure? If so, I would suggest to rename this flag from `single` to `created_by_event` (boolean as well). In my mind this would be a more precise name for the flag. Besides it is possible to figure out if a team contains only one player without having a extra column. 

Would appreciate to have you on board here. :+1:

lheilmann:
@pawelsoundso I can do the resulting refactoring, if needed. :smile:

lheilmann:
**Notes:**
- [x] Check #437 
- [x] Check #433 
- [x] Check #507

pawelsoundso:
Hey! If I understood you correctly, your assumptions about the `single` flag are correct. 

The reason for this additional column was the differentiation between single-player teams (or those teams that are created by the event) and those teams with only one participant, which are still teams and not substitutes for players. These were my intentions, why I used the additional flag and not just a method, that returns all teams that have only one player.
[...]

PR #525, opened 2018-02-04
Teams created by events should not receive team notifications

lheilmann:
See #494


app/models/team.rb
```diff
@@ -51,6 +52,10 @@ def has_member?(user)
     members.include?(user)
   end
 
+  def is_qualified_to_receive_notifications?
+    !self.created_by_event?
```
bakoe: Any reason for not using `self.created_by_user?` here? To my mind, this would increase the readability by avoiding the negation.

app/models/team_user.rb
```diff
@@ -34,6 +34,9 @@ def delete_ownership
 
   private
     def send_mail_when_created
-      TeamMailer.user_added_to_team(User.find(user_id), Team.find(team_id)).deliver_now
+      team = Team.find(team_id)
+      if team.is_qualified_to_receive_notifications?
```
bakoe, line 38: Nice!

db/schema.rb
```diff
@@ -37,8 +37,8 @@
     t.integer "points_for_win", default: 3
     t.integer "points_for_draw", default: 1
     t.integer "points_for_lose", default: 0
-    t.boolean "has_place_3_match", default: true
     t.text "image_data"
+    t.boolean "has_place_3_match", default: true
```
bakoe, line 41: You could discard/revert this change as was well the one from line 154/155. I don’t know why this always happens—the same re-ordering of existing columns in the Schema happens on my machine as well when running newly created migrations.

spec/models/team_spec.rb
```diff
@@ -102,6 +102,13 @@
     expect { team.members << user }.to change { ActionMailer::Base.deliveries.length }.by(1)
   end
 
+  it 'should not notify newly added team members when team is a single' do
```
bakoe: You could change the „single“ wording into „created by event“ here as well to avoid confusion.

bakoe (commented):
Some minor comments – however, I really like the changes. Haven’t tested in Browser though.

app/models/team.rb
lheilmann: I added `created_by_user` later... But you're definitely right, I'll change it!

spec/models/team_spec.rb
lheilmann: :+1:

app/models/team.rb
lheilmann: Well, I just recognized that `created_by_user` is a scope for `Team` and no instance variable. So `self.created_by_user?` will result in lots of errors. I will revert to the original `!self.created_by_event?` or do you have any other ideas?

app/models/user.rb
```diff
@@ -78,12 +78,12 @@ def four_different_characters
 
   has_many :team_users
   has_many :teams, through: :team_users, source: :team
-  has_many :non_single_teams, -> { where single: false }, through: :team_users, source: :team
+  has_many :teams_created_by_event, -> { where created_by_event: false }, through: :team_users, source: :team
```
niconomaa: @lheilmann can you explain this line to me?
at first sight, it seems paradoxical to me that this lines reads `where created_by_event: false`. i might be missing something here, but that might also indicate that readibility could be improved?

app/models/team.rb
jonasnoki: I would agree with @lheilmann here. The only way to use created_by_user would be to intersect with the scope... not very useful.

bakoe: Aah, okay – sorry, I didn't notice that `created_by_user` is only a scope and not an instance variable. Then it's fine to leave it at the original `!self.created_by_event?`. 👍

db/schema.rb
lheilmann: @bakoe Is there a quick and easy way to do what you recommend? Otherwise I would suggest to just let rails be weird in this case... 🤪

app/models/team.rb
niconomaa: if this is purely about readability, what you could do is use `not` instead of `!` for your negation here. the operators are slightly different in precedence, but should both work here...

db/schema.rb
bakoe: @lheilmann I can think of nothing quicker than manually adding a commit that changes `schema.rb` to be exactly the same as the one on `dev` with only the renamed field you introduced as difference. This would lead to a diff of this PR that would only contain you re-naming but not the apparent re-ordering of existing database columns.

However, I think it's pretty irrelevant – furthermore, we can't know for sure that the next person running another migration would not run into these re-ordering “issues” again. Thus, I would suggest just leaving the `schema.rb` changes in your PR as they are at the moment.

bakoe: Maybe it's even an anti-pattern to manually change the `schema.rb`, so leaving it as it is now (created by `db:migrate`) is probably the best practice.

app/models/team.rb
lheilmann: Good idea! :+1:

db/schema.rb
lheilmann: Alright!

app/models/user.rb
bakoe: You're right @tomkellog – thanks for noticing! I didn't notice before but the variable is actually named wrongly. Only the “real” teams should be displayed on the Teams index page (see #488). The current variable name implies that the opposite—the ”fake“ teams created by events—is displayed on the Teams index page (see `/views/teams/index.html.erb`).

@lheilmann Please change the variable name to something like `teams_not_created_by_event`. Thanks!

lheilmann: You're both right - thanks for noticing! Already changed to `created_by_user` and refactored.

app/models/user.rb
```diff
@@ -78,12 +78,13 @@ def four_different_characters
 
   has_many :team_users
   has_many :teams, through: :team_users, source: :team
-  has_many :non_single_teams, -> { where single: false }, through: :team_users, source: :team
+  has_many :teams_created_by_user, -> { where created_by_event: false }, through: :team_users, source: :team
+  has_many :teams_created_by_event, -> { where created_by_event: true }, through: :team_users, source: :team
```
bakoe: Is `:teams_created_by_event` used anywhere? If not, it could be removed for the moment imho.

bakoe (approved):

pawelsoundso (approved):

Issue #296, opened 2018-01-07
new event end calculation in create event form


MarcelSchmidberger:
As organizer of an event 
I want define the playing schedule dates by either entering a fixed length for the game days or by defining an end date where the event should be finished

**Acceptance Criteria**

- [ ] the end date should update itself depending on start date and game day length and predicted playing days (calculated depending on game system and max player count)
- [ ] the playing day length should update depending on the event start, end and the predicted playing days

**Mockup**
![image](https://user-images.githubusercontent.com/29160369/34924102-a809309e-f9a1-11e7-82d1-df3a76f35694.png)

**Needs Customer feedback**

PR #531, opened 2018-02-04
Bruteforce/#296/new event end calculation in create event form

CoderWombat:
connect to #296


app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -21,12 +21,65 @@ $( document ).on('turbolinks:load', function() {
...
+    }
+    function durationChanged()
+    {
+
+        var start = $("#event_startdate").val();
+        var end = $("#event_enddate").val();
+        var system = $("#league_game_mode").val();
+        var gameday_dur =$("#league_gameday_duration").val();
+        var participants = $("#league_max_teams").val();
+        if(start != "" && gameday_dur != "" && system != "" && gameday_dur != undefined && system != undefined)
```
usame-algan, line 44: Why are these two additional checks for `undefined` needed here?

CoderWombat: Since these scripts should only run on the league page and these var are undefined if its not the league page if you have a better way for checking that i would be glad to implement it

app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -21,12 +21,65 @@ $( document ).on('turbolinks:load', function() {
...
+        durationChanged();
+    }
+    function durationChanged()
+    {
+
+        var start = $("#event_startdate").val();
+        var end = $("#event_enddate").val();
+        var system = $("#league_game_mode").val();
+        var gameday_dur =$("#league_gameday_duration").val();
+        var participants = $("#league_max_teams").val();
```
usame-algan, line 43: You could extract some of this into `callChanged` since its the same in both `durationChanged` and `adaptDuration`

app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -21,12 +21,65 @@ $( document ).on('turbolinks:load', function() {
...
     });
 
-    $("#event_enddate").on("change", changeDur);
+    $("#event_enddate").on("change", adaptDuration);
 
+    $("#event_gameday_duration").on("change",durationChanged);
+
+    $("#event_game_mode").on("change",callChanged);
+
+    $("#event_max_teams").on("change",callChanged);
```
usame-algan: This id is different depending on the event type. For Leagues its `league_game_mode` etc. so the method isnt being called.

usame-algan (changes requested):

app/assets/javascripts/eventscripts/create_event.js
```diff
@@ -36,6 +87,19 @@ $( document ).on('turbolinks:load', function() {
...
+    function addDays(date, days) {
+        var result = new Date(date);
+        result.setDate(result.getDate() + parseInt(days));
+        var dd = result.getDate();
+        var mm = result.getMonth()+1;
+        var y = result.getFullYear();
+        if(mm < 10){
+            mm = "0"+ mm;
+        }
+        var formattedDate = y + '-' + mm + '-' + dd;
```
usame-algan: This `formattedDate` has a different Syntax than the other `formattedDate` further down (ymmdd vs ddmmy) which should be consistent (ddmmy)

usame-algan (changes requested):

usame-algan (approved):

Issue #511, opened 2018-02-02
Move Benachrichtigungseinstellungen to profile settings


svitlanakalancha:
Currently `Benachrichtigungseinstellungen` are placed separately on the user menu. 
Notification settings should be integrated to the `Profil bearbeiten` page.

**Acceptance criteria**
- Notification settings function as earlier
- Settings are moved to the profile settings page under separate section after `Lieblingssportarten`

birneamstiel:
also: remove notification settings page

birneamstiel:
new position: append to existing form fields, after password fields

PR #536, opened 2018-02-05
Move notification settings

birneamstiel:
Moved the notifications settings to the user's _edit profile_ page, see #511.


spec/controllers/users_controller_spec.rb
```diff
@@ -340,36 +340,4 @@
...
-    context 'given no logged in user' do
-      it 'should deny access' do
-        get :notifications, params: { id: @user.to_param }
-        expect(response).to be_unauthorized
-      end
-    end
-  end
-
-  describe 'PATCH #update_notifications' do
-    it 'should update notifications settings successfully' do
```
jonasnoki, line 361: Where did this test go?

jonasnoki (changes requested):

spec/controllers/users_controller_spec.rb
birneamstiel: Those controller actions don't existing anymore, since the updating logic is handled by the default #update action.

jonasnoki: But shouldn't there be some test that tests if you can use the #update action to change the `event_notifications_enabled` and `team_notifications_enabled` flags?

config/locales/de.yml
```diff
@@ -422,6 +414,10 @@ de:
       unhappy: "Unzufrieden"
       leave_blank_if_you_don_t_want_to_change_it: "Lass es leer, wenn du es nicht ändern möchtest"
       we_need_your_current_password_to_confirm_your_changes: "Wir benötigen dein aktuelles Passwort, um deine Änderungen zu bestätigen"
```
niconomaa, line 418: lol @ those long names

niconomaa (approved):
Hi @birneamstiel!

good work!!

4 real: didn't see anything suspicious about your code and i can confirm the settings are now in the right place after testing in browser.

i assume you talked to PO about the settings not appearing after `Lieblingssportarten` as specified in the issue?

also, i could not locally test if the mail functionality remained as it was. however, i trust you, birne. i really do.

spec/controllers/users_controller_spec.rb
niconomaa: after having an intensive chat with @jonasnoki about just that, i am joining in on the question whether this should not be tested

niconomaa (changes requested):

spec/controllers/users_controller_spec.rb
birneamstiel: @jonasnoki Yeah makes sense... I'll adjust it in a minute.

jonasnoki (approved):

niconomaa (approved):

PR #537, opened 2018-02-05
Add test for back button on events ranking page

jonasnoki:


app/views/leagues/ranking.html.erb
```diff
@@ -34,3 +34,9 @@
     <% end %>
   </tbody>
 </table>
+
+<%# this didnt work here %>
+<%# app/helpers/button_helper.rb %>
+<%= back_btn %>
+<%# this did, kind of reimplementing the feature but what do you say?%>
+<a class="btn btn-default" href="javascript:history.back()" ><%= t('helpers.links.back') %></a>
```
niconomaa: does `javascript:history.back()` not imitate the  brower's back button, which could potentially lead to a user being navigated out of our system? I know that back buttons in other places of Sportify already work this (wroooong) way.
What was the prob with `+<%= back_btn %>` ?

config/locales/en.yml
```diff
@@ -124,11 +131,6 @@ en:
           goals_against: "points against"
           goals_difference: "points difference"
           points: "league points"
-    join_team_event: "Join event %{event} as a team"
-    max_teams:
-      infinite: "unlimited"
-    overview:
-      not_available_for_leagues: 'No overview available for leagues.'
```
niconomaa, line 133: how come the number of changed lines in both translation files is not the same?

niconomaa (changes requested):

config/locales/en.yml
jonasnoki: This diff looks very strange to me too. It's somhow not showing the lines that have been moved but deleting the lines in between and inserting them before.

app/views/leagues/ranking.html.erb
jonasnoki: The  `<%= back_btn %>`  is actually doing the same thing but when this view is loaded the back button points to the view itself. What do you think the back button should be doing? Or maybe thats a question for our PO. @svitlanakalancha What do you think?

birneamstiel: Another suggestion: how about adding a specific target for the back button? Something like this: `<%= link_to t('helpers.links.back'), league_path(@event), class: 'btn btn-default' %>`

bakoe: To my mind, the back button should always simply point to the corresponding league's page. This would imply that ”Move back“ means something like ”Move one layer up in the hierarchical layering of our app” aka “Move up”. I think this is a common practice for back buttons on those kinds of applications – however, asking our PO sounds like a good idea. 👍 

**EDIT:** @birneamstiel's comment was not yet displayed when I wrote my comment. I actually mean exactly what @birneamstiel proposed. 🙂

db/schema.rb
```diff
@@ -31,13 +31,13 @@
...
     t.integer "min_players_per_team"
     t.integer "max_players_per_team"
+    t.text "image_data"
     t.integer "matchtype"
     t.integer "bestof_length", default: 1
     t.integer "game_winrule"
     t.integer "points_for_win", default: 3
     t.integer "points_for_draw", default: 1
     t.integer "points_for_lose", default: 0
-    t.text "image_data"
```
birneamstiel: I'd rather not have this in here... But maybe it will resolve anyways when merging a newer dev.

config/locales/en.yml
birneamstiel: Why were there even changes made to translation files in this PR?

app/views/leagues/ranking.html.erb
niconomaa: yup, i agree. that'd then lead to the back button actually just being a conventional link in disguise, i would assume.

svitlanakalancha: The back button should lead to the correspondent event page as the story says. I agree with @bakoe opinion. We could rename it to 'Zurück zur Liga' or 'Zurück zum Turnier', but I don't think it is necessary. User can open the table only from event page, so it may be understandable, what happens by clicking 'Zurück' :)

config/locales/en.yml
jonasnoki: Because I moved the ranking view to the leagues folder.

birneamstiel:
I'd consider renaming this PR to something like "Add test for back button on ranking page" since the back button was apparently added by @ErikBrendel already in a3a07b31398215cebc299ea263920f8ef1a8b5b3 :-) But otherwise I think we're good to go ;-)

birneamstiel (approved):

jonasnoki:
Yes, I just merged the dev into this branch which created conflicts in the view ranking file. A very stable solution for the back_btn had appeared. Kind of sad to have worked 3 hours for basically nothing because of duplicated issues.  I guess that is what this lecture is about.
@tomkellog would you accept the changes so we can merge this awesome test, plzzzz?

niconomaa (approved):

Issue #524, opened 2018-02-03
Remove unnecessary buttons from playing schedule


MarcelSchmidberger:
`Spielplan` page contains unnecessary buttons which confuse a user. This story only affects the design of the page, not its functionality.

**Acceptance Criteria**

- [ ] remove `Löschen` button
- [ ] remove `Bearbeiten` button
- [ ] rename `Anzeigen` button to `Ergebnisse eintragen`

![image](https://user-images.githubusercontent.com/29160369/35772363-0d89db44-093d-11e8-8916-df62f8fcac72.png)

birneamstiel:
In consultation with @svitlanakalancha: changes affect both `Tournaments` and `Leagues`

PR #539, opened 2018-02-05
Remove unnecessary buttons from schedule page

birneamstiel:
Task was to delete two buttons (for tournament only one) and rename the remaining one, see #524.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -131,6 +131,8 @@ en:
       infinite: "unlimited"
     overview:
       not_available_for_leagues: 'No overview available for leagues.'
+    schedule:
+      insert_results: "Insert result"
 
   teams:
       show:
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -131,6 +131,8 @@ en:
       infinite: "unlimited"
     overview:
       not_available_for_leagues: 'No overview available for leagues.'
+    schedule:
+      insert_results: "Insert result"
```
bakoe, line 135: Have you used the singular (result) here on purpose? The key name (insert_results) made me think that the localization should be „Insert results“

bakoe (approved):
Only one apparent typo. Apart from that: LGTM! Haven’t tested in Browser though – due to the simplicity of the changes and the added test, I am pretty sure it’s working correctly. 😉

conradlempert (approved):
nicer dicer

Issue #490, opened 2018-01-31
Invite new player pop-up doesn't refresh values


svitlanakalancha:
**Actual behavior**
- Create new team and click `+` button
- Pop-up form to invite new player is opened
- Put some text in the field and click **Abbrechen**
- Click `+` button again - > your text remains in the field

**Correct behavior**
- Text should be deleted after closing the pop-up form

**Note**
Please rework black background, it should not be shown.

<img width="1007" alt="sportify" src="https://user-images.githubusercontent.com/27727717/35611493-74dc7520-0665-11e8-9ddd-9ee0cb7dc5c3.png">

svitlanakalancha:
@tomkellog please check if it is possible to remove white background completely. Also it would be great to place buttons in line so that they fit the pop-up form

![screen shot 2018-02-06 at 12 04 28](https://user-images.githubusercontent.com/27727717/35856310-e9ce5c66-0b35-11e8-9b2d-d151acc18155.png)

PR #545, opened 2018-02-05
Kickerz/refresh text in invite modal

niconomaa:
Adds the necassary script to clear the modal when it is closed. Also removes the black backdrop of the modal as specified in the issues's 'note'. ;)

 
After some trying, I omitted testing, since,

in short: JS testing in Capybara is a bitch and we never really test view side js anyways.

in full: in order to test JS in our tests we'd need to switch the `Capybara driver` to `webkit` for the test only, which would require installing the necessary gem, which is, it seems, [not so easy](https://github.com/thoughtbot/capybara-webkit/wiki/Installing-Qt-and-compiling-capybara-webkit).
If someone still insists on testing the script, I'd be glad to receive some input on how go about it either using the problematic approach I mentioned above or another one.

Make sure to use the whitespace insensitive GitHub view. :)

### DIFF
--- a/app/assets/stylesheets/teams.css
+++ b/app/assets/stylesheets/teams.css
@@ -3,6 +3,10 @@
   They will automatically be included in application.css.
 */
 
+.modal-backdrop.fade.in {
+  opacity: 0;
+}
+
 label.required:after {
   content: "*";
   vertical-align: 0.5em;
--- a/app/assets/javascripts/teamscripts/teams.js
+++ b/app/assets/javascripts/teamscripts/teams.js
@@ -1,22 +1,29 @@
 $(document).ready(function(){
-// Place all the behaviors and hooks related to the matching controller here.
-// All this logic will automatically be available in application.js.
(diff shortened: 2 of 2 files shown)


lheilmann (approved):
LGTM code-wise - tested in browser, too! :+1:

app/assets/stylesheets/teams.css
```diff
@@ -3,6 +3,10 @@
   They will automatically be included in application.css.
 */
 
+.modal-backdrop.fade.in {
```
birneamstiel, line 6: just out of curiosity: what's this change for?

lheilmann:
I was wondering why there is still a white box behind the modal. Besides the buttons do not look appropriate (see screenshot).

screen shot 2018-02-06 at 10 21 29

But these issues do not seem to be in scope of #490. Is this intended to remain like that @svitlanakalancha, just to be sure?

birneamstiel (approved):

lheilmann:
https://getbootstrap.com/docs/3.3/javascript/#modals-sizes

PR #548, opened 2018-02-06
Move “Spiel um Platz 3” checkbox on Create tournament page into the first section

bakoe:
The addition of the CSS class `order-1` places the form-group of the has_place_3_match checkbox within the first section of the Create tournament form.

Placing the corresponding form-group on the beginning of `_Tournament.html.erb` (and thus before the form-group with `order-4`) is not necessary, but increases the readability/maintainability of the .erb file to my mind.


niconomaa (approved):
straight forward

lheilmann (approved):
LGTM :+1:

PR #551, opened 2018-02-06
Kickerz/#481/hide match datechange button from unauthorized users

niconomaa:
Hides the datechange button if the user can't edit leagues.


conradlempert (approved):
code is ganz nais, da gibts halt noch nen unhappy cucumber test bro

spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -25,11 +25,19 @@
...
-    it 'has edit button for gameday dates' do
+    it 'has edit button for gameday dates for signed in users' do
+      @user = FactoryBot.create :user
+      @league.owner_id = @user.id
+      sign_in @user
       render
       expect(rendered).to have_selector(:link_or_button, t('events.schedule.edit_date'))
     end
 
+    it 'does not have edit button for gameday dates for signed ou users' do
```
lheilmann: Typo: `out`

spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -25,11 +25,19 @@
...
+      render
+      expect(rendered).not_to have_selector(:link_or_button, t('events.schedule.edit_date'))
+    end
+
```
lheilmann: Consider another test case for an user, who is signed in but has no rights to see this particular button (e.g. no event organizer rights) - this would be an ability test though. Is there already one that covers such behaviour?

lheilmann (changes requested):
Just a typo and an idea for another test, if not present yet. :+1:

spec/views/events/schedule.html.erb_spec.rb
niconomaa: Good idea, but I don't think this must be an Ability test necessarily (the corresponeding `edit` is tested seperately). I am, however, adding a test with a signed in user with no rights to edit!

spec/views/events/schedule.html.erb_spec.rb
```diff
@@ -27,9 +27,28 @@
...
-    it 'has edit button for gameday dates' do
-      render
-      expect(rendered).to have_selector(:link_or_button, t('events.schedule.edit_date'))
+    describe 'authorization for edit date button' do
+      before(:each) do
+        @user = FactoryBot.create(:user)
+        sign_in @user
+      end
+
+      it 'has edit button for gameday dates for signed in users' do
```
lheilmann: Consider using the following as description: `has edit button for gameday dates for signed in users with rights as organizer of the league`

lheilmann (approved):

PR #557, opened 2018-02-06
Flexibus/#380/events view as tiles

### LINKED ISSUE #380: Events view as tiles
**This depends on #379 for tile component**
**#374 depends on this for event logic**

New Layout for "Events" Page:

If the event has a event-picture, then tile has this picture. If not, then we use 1 of 2 default pictures (they can be easily found in Flexibus channel)

If width for content is <=800px, then it's only 2 blocks in a row.

- 'Deine Events' = I'm a member, my team is a member, I created the event
- 'Deine Events' is not shown when there is no data

jcremerius:
Closes #380.


jkbe:
As discussed in the last meeting: Please move the corresponding issue to "in review" yourself on Waffle.io (or simply write "Closes #<issuenumber>" in the PR description) so that we have a correct overview of the state of tickets for Kanban! @jcremerius

app/models/event.rb
```diff
@@ -186,11 +186,10 @@ def human_game_mode
...
-    if self.type == 'Rankinglist'
-      "#{I18n.t('events.index.registration_until')}: #{self.deadline} <br> #{I18n.t('events.index.start_date')}: #{self.startdate}"
-    else
-      "#{I18n.t('events.index.max_players')}: #{self.max_teams}"
-    end
+    registration_until = "#{I18n.t('events.index.registration_until')}: #{self.deadline}" if self.deadline.present?
+    start_date = "#{I18n.t('events.index.start_date')}: #{self.startdate}" if self.startdate.present?
+    return ("#{registration_until} <br> #{start_date}").html_safe if registration_until.present?
+    "#{I18n.t('events.index.max_players')}: #{self.max_teams}" if self.max_teams.present?
   end
```
Jeilef, line 193: does this method really have to be in the Events model?

jcremerius: Where would you place it?

app/assets/stylesheets/tile.css
```diff
@@ -14,6 +14,18 @@
...
         width: 300px;
     }
+
+    .tile-title {
+        width: 300px;
+    }
+}
+
+
+@media screen and (max-width : 799px) and (min-width: 481px) {
```
jkbe: What do we have this block for? The condition selects a subset of elements of the `@media`'s condition above and the CSS values are the same. Remove?

jkbe (changes requested):

jkbe (dismissed):
LGTM but...

jkbe (approved):

PR #558, opened 2018-02-06
Flexibus/#381/team page design

### LINKED ISSUE #381: Team Page Design

deeps96:
Closes #381. Fixes max-width issue described in #379.


db/schema.rb
```diff
@@ -31,13 +31,13 @@
     t.integer "selection_type", default: 0, null: false
     t.integer "min_players_per_team"
     t.integer "max_players_per_team"
+    t.text "image_data"
```
jkbe: Please undo the changes you made in this file as they're only adding noise.

jkbe (changes requested):
@zhamanakov in #375 @MerlindlH and his team (@ErikBrendel @Oluwoye) added back buttons to all pages. Are we sure that we want to remove the button from the show team page?

jkbe:
@deeps96 ~Is there any specification for the last commit you made?~ 
Edit: Spotted the fix now, makes sense to do that. 👍  ~What's different to before? Can't spot anything on the checked out branch locally.~
What about testing that the header is not visible if not signed in?

##Response 1
Will add test - DONE

jkbe:
LGTM except for back button question, modified schema.rb and missing test.

db/schema.rb
deeps96: Whoah where did that came from 😮 - Will fix

jkbe: Happens sometimes when running migrations.

deeps96: Done

deeps96:
Implemented all changes requested by @jak-ing

jkbe:
Talked to @zhamanakov and he said he's already agreed with @MerlindlH on removing the backbutton from the team page. Merging soon.

jkbe (approved):

PR #560, opened 2018-02-07
Make the invite user by email modal on Show team page look pretty

bakoe:
See #490

image


app/views/teams/show.html.erb
```diff
@@ -41,6 +41,10 @@
                 $('.only-show-when-participants-selected').hide();
             }
         });
+
+        $('#invite_user_by_email_modal').modal({
+            complete: function() { $(".modal-body #email").val(""); } // Clear email field when modal closes (see #490)
+        });
     });
```
niconomaa, line 48: why here?

niconomaa (approved):

app/views/teams/show.html.erb
bakoe: Because `app/assets/javascripts/teamscripts/teams.js` seems more Controller-related to me (especially regarding the following comment in the file)

```javascript
     // Place all the behaviors and hooks related to the matching controller here.
     // All this logic will automatically be available in application.js.
```

and `app/views/teams/show.html.erb` already contains a lot of JavaScript code related to the actual Show team page:

```javascript
$('.check_all')
…
$('.uncheck_all')
…
$('.send_email_to_multiple_members')
…
$('.only-show-when-participants-selected')
…
$('.participant :checkbox')
```

birneamstiel (approved):
Looks good to me, tested in browser 👍

PR #563, opened 2018-02-07
Swteam/#369/profile validations

### LINKED ISSUE #369: profile data integrity
*note: this story was split up in order to make it easier to complete the parts*

new Story: 
**As an** admin
**I want** my users **to** enter their data in a sensible/useful way
**In order to** not confuse other users

## Acceptance criteria:
 * [x] all Profile information should be validated:
    * [x] telephone numbers should be in telephone number format (with + at the beginning allowed as well)
     * [x] the current day is not in the future... (1)

### pictures:
(1)
 image

fpottbaecker:
This will close #369.


spec/models/user_spec.rb
```diff
@@ -47,21 +47,41 @@
...
+    it 'should be valid with a fully specified phone number' do
+      expect(FactoryBot.build(:user, telephone_number: "+00491766563")).to be_valid
+    end
+
+    it 'should be valid with a same country phone number' do
+      expect(FactoryBot.build(:user, telephone_number: "01731557326")).to be_valid
+    end
+
+    it 'should not be valid when the number contains spaces' do
+      expect(FactoryBot.build(:user, telephone_number: "0049 177 499 372")).not_to be_valid
```
ErikBrendel, line 69: wouldn't it be helpful if we just stripped the spaces somewhere, so that the user does not get errors when entering space-numbers?

spec/models/user_spec.rb
```diff
@@ -47,21 +47,41 @@
...
+    it 'should be valid with a past birthday' do
+      expect(FactoryBot.build(:user, birthday: Date.today - 20.years)).to be_valid
+    end
+
+    it 'should be valid with a birthday today' do
+      expect(FactoryBot.build(:user, birthday: Date.today)).to be_valid
+    end
+
+    it 'should not be valid with a future birthday' do
+      expect(FactoryBot.build(:user, birthday: Date.tomorrow))
```
ErikBrendel: yep, expect that user!

spec/models/user_spec.rb
ErikBrendel: ( https://stackoverflow.com/questions/28621354/strip-phone-number-input-before-validation-in-rails ) maybe?

Markus-Brand (approved):

spec/models/user_spec.rb
fpottbaecker: PO says: "It's ok, it's a phone number, and does not have that high a priority"

spec/models/user_spec.rb
fpottbaecker: NOBODY EXPECTS THE SPANISH INQUISITION!

